PR #1, opened 2025-06-16
Remote GSM

efecarranza:
# Changelog

Add RemoteGSM Implementation
Abstract methods on GSM

### DIFF
--- a/.github/workflows/certora-gho.yml
+++ b/.github/workflows/certora-gho.yml
@@ -61,3 +61,4 @@ jobs:
           install-java: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CERTORAKEY: ${{ secrets.CERTORAKEY }}
--- a/.github/workflows/certora-gsm-4626.yml
+++ b/.github/workflows/certora-gsm-4626.yml
@@ -59,3 +59,4 @@ jobs:
           install-java: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CERTORAKEY: ${{ secrets.CERTORAKEY }}
--- a/.github/workflows/certora-gsm.yml
+++ b/.github/workflows/certora-gsm.yml
@@ -57,3 +57,4 @@ jobs:
           install-java: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CERTORAKEY: ${{ secrets.CERTORAKEY }}
--- a/.github/workflows/certora-steward.yml
+++ b/.github/workflows/certora-steward.yml
@@ -49,3 +49,4 @@ jobs:
           install-java: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CERTORAKEY: ${{ secrets.CERTORAKEY }}


src/contracts/facilitators/gsm/GhoReserve.sol
```diff
@@ -0,0 +1,131 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import {IERC20} from 'aave-v3-origin/contracts/dependencies/openzeppelin/contracts/IERC20.sol';
+import {Ownable} from '@openzeppelin/contracts/access/Ownable.sol';
+import {EnumerableSet} from '@openzeppelin/contracts/utils/structs/EnumerableSet.sol';
+import {VersionedInitializable} from 'aave-v3-origin/contracts/misc/aave-upgradeability/VersionedInitializable.sol';
```
miguelmtzinf, line 7:
```suggestion
import {Ownable} from '@openzeppelin/contracts/access/Ownable.sol';
import {EnumerableSet} from '@openzeppelin/contracts/utils/structs/EnumerableSet.sol';
import {IERC20} from 'aave-v3-origin/contracts/dependencies/openzeppelin/contracts/IERC20.sol';
import {VersionedInitializable} from 'aave-v3-origin/contracts/misc/aave-upgradeability/VersionedInitializable.sol';
```

src/contracts/facilitators/gsm/OwnableFacilitator.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import {Ownable} from '@openzeppelin/contracts/access/Ownable.sol';
+import {IOwnableFacilitator} from 'src/contracts/facilitators/gsm/interfaces/IOwnableFacilitator.sol';
+import {IGhoToken} from 'src/contracts/gho/interfaces/IGhoToken.sol';
```
miguelmtzinf, line 6:
```suggestion
import {IGhoToken} from 'src/contracts/gho/interfaces/IGhoToken.sol';
import {IOwnableFacilitator} from 'src/contracts/facilitators/gsm/interfaces/IOwnableFacilitator.sol';
```

src/contracts/facilitators/gsm/GhoReserve.sol
```diff
@@ -47,6 +47,8 @@ contract GhoReserve is Ownable, VersionedInitializable, IGhoReserve {
 
   /// @inheritdoc IGhoReserve
   function use(uint256 amount) external {
+    require(_entities.contains(msg.sender), "NOT_ENTITY");
```
miguelmtzinf, line 50: is this check actually needed? same in `restore`
- when removing the entity, we check used and limit is 0.
- when adding an entity, it can start with 0 limit.
- use check limit, so only existent entities with non-zero limit are allowed
- restore check used, so only existent entities with non-zero used are allowed

miguelmtzinf: apart from this, we are doing a similar check in `setLimit` with different error. We should be consistent
```
  function setLimit(address entity, uint256 limit) external onlyOwner {
    require(_entities.contains(entity), 'ENTITY_DOES_NOT_EXIST');
```

PR #5, opened 2025-07-18
sGHO - erc4626 vault

Luigy-Lemon:
# sGHO - Savings GHO Vault

## About

sGHO is an [EIP-4626](https://eips.ethereum.org/EIPS/eip-4626) vault that allows users to earn yield on their GHO tokens. The vault automatically accrues and distributes yield to depositors through internal accounting, with all logic self-contained in the sGHO contract.

## Features

- **Full [EIP-4626](https://eips.ethereum.org/EIPS/eip-4626) compatibility.** sGHO implements all standard ERC-4626 functions for deposits, withdrawals, and share calculations.
- **Automatic yield accrual.** Yield is accrued and distributed internally during vault operations, compounding linearly between state updates (e.g., deposit, withdraw, mint, redeem).
- **Role-based access control.** Uses Aave's ACLManager for managing permissions:
  - `YIELD_MANAGER_ROLE`: Can set the target APR and supply cap
  - `FUNDS_ADMIN_ROLE`: Can rescue non-GHO tokens in emergencies
- **Permit support.** Users can approve sGHO to spend their GHO tokens using EIP-2612 permits, enabling gasless approvals.
- **No ETH acceptance.** The contract rejects direct ETH transfers.
- **Supply cap.** The vault enforces a maximum cap on total GHO that can be deposited.

## Architecture

The system consists of a single main contract:

**sGHO.sol**: The main vault contract that:
- Implements ERC4626 for standard vault operations
- Handles deposits and withdrawals with automatic yield accrual
- Manages internal asset accounting with a yield index
- Integrates with Aave's ACLManager for role management

### sGHO.sol
- Inherits from:
  - `ERC4626Upgradeable` (OpenZeppelin) - For vault functionality
  - `ERC20PermitUpgradeable` (OpenZeppelin) - For permit functionality
  - `Initializable` (OpenZeppelin) - For initialization pattern
  - `IsGHO` (Custom interface) - Contract interface
- Uses OpenZeppelin interfaces:
  - `IERC20` - For token operations
  - `IAccessControl` - For role management
- Uses Aave libraries:
  - `WadRayMath` - For precise mathematical calculations

## Yield Accrual and Distribution

Yield is accrued and distributed automatically during vault operations. The yield compounds linearly between state updates (e.g., deposit, withdraw, mint, redeem) and is tracked via a `yieldIndex`.

### Key Components:
- **Target Rate**: Annual percentage rate in basis points (e.g., 1000 = 10%)
- **Yield Index**: Tracks the cumulative yield multiplier (in ray, 1e27 scale)
- **Last Update**: Timestamp of the last yield accrual
- **Supply Cap**: Maximum GHO that can be deposited in the vault

### Yield Accrual Process:
1. When a vault operation occurs (deposit, withdraw, mint, redeem), `_updateYieldIndex()` is called.
2. The yield index is updated based on the elapsed time and the current target rate.
3. Share/asset conversions use the up-to-date yield index for accurate accounting.

#### Example (from contract):
```solidity
function _getCurrentYieldIndex() internal view returns (uint256) {
  if (targetRate == 0) return yieldIndex;
  uint256 timeSinceLastUpdate = block.timestamp - lastUpdate;
  if (timeSinceLastUpdate == 0) return yieldIndex;
  uint256 annualRateRay = uint256(targetRate).rayDiv(10000);
  uint256 ratePerSecond = annualRateRay.rayDiv(365 days);
  uint256 accumulatedRate = ratePerSecond.rayMul(timeSinceLastUpdate);
  uint256 growthFactor = WadRayMath.RAY + accumulatedRate;
  return yieldIndex.rayMul(growthFactor);
}
```

## Security Considerations

- **Role-based access control** through Aave's ACLManager
- **Token rescue mechanism** for handling stuck non-GHO tokens (GHO cannot be rescued)
- **No ETH acceptance** to prevent accidental ETH deposits
- **Initialization pattern** prevents re-initialization attacks
- **Supply cap** to limit total GHO in the vault

## Limitations

- Target rate and supply cap can only be modified by accounts with the YIELD_MANAGER_ROLE
- The system requires GHO tokens to be properly configured and accessible
- Withdrawals and redemptions are limited by the contract's actual GHO balance (no explicit buffer logic)
- Yield is accrued automatically during vault operations, not on-demand
- **GHO Balance Dependency**: While yield accrues based on time and target rate, actual withdrawals are limited by the contract's GHO balance. Users can check available withdrawal capacity using `maxWithdraw()` and `maxRedeem()` functions, which return the minimum of the user's share value and the contract's actual GHO balance.

## Shortfall Risk and Capitalization

**Important**: sGHO operates on a first-come, first-served basis for withdrawals. If the contract's GHO balance falls below the total value of shares (totalAssets), some depositors may be unable to withdraw their full balance until additional GHO is provided to the contract.

This creates an expectation that the AAVE DAO will maintain sGHO sufficiently capitalized above the value of totalAssets to ensure all depositors can withdraw their funds when desired. The DAO should monitor the contract's GHO balance and replenish it as needed to maintain adequate liquidity for withdrawals.

Users should be aware that during periods of insufficient GHO balance, withdrawal requests may be partially filled or rejected entirely, depending on the available GHO in the contract at the time of the withdrawal attempt.

### Depositing GHO
```solidity
// Deposit GHO into the vault
sgho.deposit(amount, receiver);

// Deposit with permit (gasless approval)
sgho.permit(owner, spender, value, deadline, v, r, s);
sgho.deposit(amount, receiver);

// Mint shares for a specific amount of GHO
sgho.mint(shares, receiver);
```

### Withdrawing GHO
```solidity
// Withdraw GHO from the vault
sgho.withdraw(assets, receiver, owner);

// Redeem shares for GHO
sgho.redeem(shares, receiver, owner);
```

### Managing Yield and Configuration
```solidity
// Set target rate (YIELD_MANAGER_ROLE only)
sgho.setTargetRate(1000); // 10% APR

// Set supply cap (YIELD_MANAGER_ROLE only)
sgho.setSupplyCap(newCap);

// View current vault APR
uint256 apr = sgho.vaultAPR();

// Rescue non-GHO tokens in emergency (FUNDS_ADMIN_ROLE only)
sgho.rescueERC20(tokenAddress, recipient, amount);
```

### Permit Usage (Gasless Approvals)
```solidity
// Standard permit
sgho.permit(owner, spender, value, deadline, v, r, s);
```

## Important Notes

- **Supply Cap**: The vault enforces a maximum cap on total GHO that can be deposited, providing a mechanism to control the pool size.
- **Automatic Yield**: Yield is accrued and distributed automatically during deposit/withdrawal/mint/redeem operations.
- **No External Yield Management**: All yield logic is internal; there is no separate yield manager contract.
- **Self-contained Architecture**: The system is fully contained within the sGHO contract.
- **Composable Design**: sGHO implements the ERC4626 standard, making it composable with DeFi protocols that support this standard (lending protocols, yield aggregators, etc.).
- **Cross-chain Deployment**: The contract can be deployed on any EVM-compatible blockchain, enabling GHO yield opportunities across multiple networks.

---

This README is up-to-date with the current implementation of `sGHO.sol` as of this version. Please refer to the contract for the most precise technical details.

### DIFF
--- a/src/contracts/sgho/interfaces/IsGHO.sol
+++ b/src/contracts/sgho/interfaces/IsGHO.sol
@@ -1,0 +1,198 @@
+// SPDX-License-Identifier: agpl-3
+pragma solidity ^0.8.19;
+
+/**
+ * @title IsGHO Interface
+ * @notice Interface for the sGHO contract, which is an ERC4626 vault for GHO tokens.
+ * @dev This interface combines functionalities from ERC4626 for a tokenized vault,
+ * ERC20Permit for gas-less approvals, and includes custom logic for yield generation and administrative roles.
+ */
+interface IsGHO {
+  // --- Custom Errors ---
+
+  /**
+   * @notice Thrown when an invalid signature is provided.
+   */
+  error InvalidSignature();
+
+  /**
+   * @notice Thrown when a direct ETH transfer is attempted.
+   */
+  error NoEthAllowed();
+
+  /**
+   * @notice Thrown when a function is called by an address that does not have the FUNDS_ADMIN role.
+   */
+  error OnlyFundsAdmin();
+
+  /**
+   * @notice Thrown when a function is called by an address that does not have the YIELD_MANAGER role.
+   */
+  error OnlyYieldManager();
+
+  /**
+   * @notice Thrown when an attempt is made to rescue the underlying GHO token.
+   */
+  error CannotRescueGHO();
+
+  /**
+   * @notice Thrown if the target rate is set to a value greater than the max rate.
+   */
+  error RateMustBeLessThanMaxRate();
+
+  /**
+   * @notice Thrown when a deposit or mint would exceed the total supply cap.
+   */
+  error SupplyCapExceeded();
+
+  // --- Events ---
+
+  /**
+   * @notice Emitted when the target rate is updated.
+   * @param newRate The new target rate.
+   */
+  event TargetRateUpdated(uint256 newRate);
+
+  /**
+   * @notice Emitted when ERC20 tokens are rescued from the contract.
+   * @param caller The address that initiated the rescue operation.
+   * @param token The address of the rescued ERC20 token.
+   * @param to The recipient address of the rescued tokens.
+   * @param amount The amount of tokens rescued.
+   */
+  event ERC20Rescued(
+    address indexed caller,
+    address indexed token,
+    address indexed to,
+    uint256 amount
+  );
+
+  // --- State Variables (as view functions) ---
+
+  /**
+   * @notice Returns the address of the GHO token used as the underlying asset in the vault.
+   * @return The address of the GHO token.
+   */
+  function gho() external view returns (address);
+
+  /**
+   * @notice Returns the total supply cap of the vault.
+   * @return The total supply cap.
+   */
+  function supplyCap() external view returns (uint256);
+
+  /**
+   * @notice Returns the chain ID of the network where the contract is deployed.
+   * @dev This is used for EIP-712 signature validation to prevent replay attacks across different chains.
+   * @return The chain ID.
+   */
+  function deploymentChainId() external view returns (uint256);
+
+  /**
+   * @notice Returns the current yield index, representing the accumulated yield.
+   * @dev This index is used to calculate the value of sGHO in terms of GHO.
+   * @return The current yield index.
+   */
+  function yieldIndex() external view returns (uint256);
+
+  /**
+   * @notice Returns the current target annual percentage rate (APR) for yield generation.
+   * @dev The rate is expressed in basis points (1% = 100).
+   * @return The target rate in basis points.
+   */
+  function targetRate() external view returns (uint256);
+
+  /**
+   * @notice Returns the timestamp of the last time the yield index was updated.
+   * @return The Unix timestamp of the last update.
+   */
+  function lastUpdate() external view returns (uint256);
+
+  /**
+   * @notice Returns the role identifier for the Funds Admin.
+   * @dev This role has permissions to manage funds, such as rescuing tokens.
+   * @return The keccak256 hash of "FUNDS_ADMIN_ROLE".
+   */
+  function FUNDS_ADMIN_ROLE() external view returns (bytes32);
+
+  /**
+   * @notice Returns the role identifier for the Yield Manager.
+   * @dev This role has permissions to update the target rate.
+   * @return The keccak256 hash of "YIELD_MANAGER_ROLE".
+   */
+  function YIELD_MANAGER_ROLE() external view returns (bytes32);
+
+  // --- Functions ---
+
+  // Note: Standard ERC4626 functions (asset, totalAssets, convertToShares, convertToAssets,
+  // maxDeposit, previewDeposit, deposit, maxMint, previewMint, mint, maxWithdraw, previewWithdraw,
+  // withdraw, maxRedeem, previewRedeem, redeem) and ERC20 functions (name, symbol, decimals,
+  // totalSupply, balanceOf, transfer, allowance, approve, transferFrom) are inherited via IERC4626.
+
+  // Note: Standard ERC20Permit functions (permit, nonces, DOMAIN_SEPARATOR) are inherited via IERC20Permit.
+
+  /**
+   * @notice Initializes the sGHO contract.
+   * @dev This function can only be called once. It sets up initial roles and configurations.
+   * While the function is marked as `payable`, it is designed to reject any attached Ether value.
+   */
+  function initialize(
+    address gho_,
+    address aclManager_,
+    uint256 maxTargetRate_,
+    uint256 supplyCap_
+  ) external payable;
+
+  /**
+   * @notice Overload of the standard ERC20Permit `permit` function.
+   * @dev This version accepts the v, r, and s components of the signature directly,
+   * which can be useful for platforms that do not handle the single `bytes` signature format.
+   * @param owner The owner of the tokens.
+   * @param spender The address to grant allowance to.
+   * @param value The amount of allowance to grant.
+   * @param deadline The timestamp after which the permit is invalid.
+   * @param v The recovery ID of the signature.
+   * @param r The R component of the signature.
+   * @param s The S component of the signature.
+   */
+  function permit(
+    address owner,
+    address spender,
+    uint256 value,
+    uint256 deadline,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external;
+
+  /**
+   * @notice Sets the target rate for yield generation.
+   * @dev This function can only be called by an address with the YIELD_MANAGER role.
+   * The new rate must be less than 50% (5000 basis points).
+   * @param newRate The new target rate in basis points (e.g., 1000 for 10%).
+   */
+  function setTargetRate(uint256 newRate) external;
+
+  /**
+   * @notice Calculates and returns the current vault Annual Percentage Rate (APR).
+   * @return The current vault APR, in basis points (1% = 100).
+   */
+  function vaultAPR() external view returns (uint256);
+
+  /**
+   * @notice Rescues ERC20 tokens that have been accidentally sent to this contract.
+   * @dev This function can only be called by an address with the FUNDS_ADMIN role.
+   * It prevents the rescue of the underlying GHO token to protect the vault's assets.
+   * @param erc20Token The address of the ERC20 token to rescue.
+   * @param to The address where the rescued tokens will be sent.
+   * @param amount The amount of tokens to rescue.
+   */
+  function rescueERC20(address erc20Token, address to, uint256 amount) external;
+
+  // --- Events ---
+
+  /**
+   * @notice The receive function is implemented to reject direct Ether transfers to the contract.
+   */
+  receive() external payable;
+}
--- a/src/contracts/sgho/interfaces/IsGHO.sol
+++ b/src/contracts/sgho/interfaces/IsGHO.sol
@@ -88,12 +88,12 @@ interface IsGHO {
   function supplyCap() external view returns (uint256);
 
   /**
-   * @notice Returns the chain ID of the network where the contract is deployed.
-   * @dev This is used for EIP-712 signature validation to prevent replay attacks across different chains.
-   * @return The chain ID.
+   * @notice Returns the maximum safe rate for the vault.
+   * @return The maximum safe rate.
    */
-  function deploymentChainId() external view returns (uint256);
+  function MAX_SAFE_RATE() external view returns (uint16);
 
+ 
   /**
    * @notice Returns the current yield index, representing the accumulated yield.
    * @dev This index is used to calculate the value of sGHO in terms of GHO.
@@ -106,7 +106,7 @@ interface IsGHO {
    * @dev The rate is expressed in basis points (1% = 100).
    * @return The target rate in basis points.
    */
-  function targetRate() external view returns (uint256);
+  function targetRate() external view returns (uint16);
 
   /**
    * @notice Returns the timestamp of the last time the yield index was updated.
@@ -145,7 +145,6 @@ interface IsGHO {
   function initialize(
     address gho_,
     address aclManager_,
-    uint256 maxTargetRate_,
     uint256 supplyCap_
   ) external payable;
 
@@ -177,7 +176,7 @@ interface IsGHO {
    * The new rate must be less than 50% (5000 basis points).
    * @param newRate The new target rate in basis points (e.g., 1000 for 10%).
    */
-  function setTargetRate(uint256 newRate) external;
+  function setTargetRate(uint16 newRate) external;
 
   /**
    * @notice Sets the supply cap for the vault.
--- a/src/contracts/sgho/interfaces/IsGHO.sol
+++ b/src/contracts/sgho/interfaces/IsGHO.sol
@@ -11,11 +11,6 @@ interface IsGHO {
   // --- Custom Errors ---
 
   /**
-   * @notice Thrown when an invalid signature is provided.
-   */
-  error InvalidSignature();
-
-  /**
    * @notice Thrown when a direct ETH transfer is attempted.
    */
   error NoEthAllowed();
@@ -45,6 +40,11 @@ interface IsGHO {
    */
   error SupplyCapExceeded();
 
+  /**
+   * @notice Thrown when the supply cap is set to a value less than the total assets.
+   */
+  error SupplyCapMustBeGreaterThanTotalAssets();
+
   // --- Events ---
 
   /**
@@ -52,6 +52,12 @@ interface IsGHO {
    * @param newRate The new target rate.
    */
   event TargetRateUpdated(uint256 newRate);
+
+  /**
+   * @notice Emitted when the supply cap is updated.
+   * @param newSupplyCap The new supply cap.
+   */
+  event SupplyCapUpdated(uint256 newSupplyCap);
 
   /**
    * @notice Emitted when ERC20 tokens are rescued from the contract.
@@ -172,6 +178,13 @@ interface IsGHO {
    * @param newRate The new target rate in basis points (e.g., 1000 for 10%).
    */
   function setTargetRate(uint256 newRate) external;
+
+  /**
+   * @notice Sets the supply cap for the vault.
+   * @dev This function can only be called by an address with the YIELD_MANAGER role.
+   * @param newSupplyCap The new supply cap.
+   */
+  function setSupplyCap(uint256 newSupplyCap) external;
 
   /**
    * @notice Calculates and returns the current vault Annual Percentage Rate (APR).
--- a/src/contracts/sgho/interfaces/IsGHO.sol
+++ b/src/contracts/sgho/interfaces/IsGHO.sol
@@ -45,6 +45,16 @@ interface IsGHO {
    */
   error SupplyCapMustBeGreaterThanTotalAssets();
 
+  /**
+   * @notice Thrown when a zero address is provided for a critical parameter during initialization.
+   */
+  error ZeroAddressNotAllowed();
+
+  /**
+   * @notice Thrown when an ERC20 transfer fails during rescue operation.
+   */
+  error TransferFailed();
+
   // --- Events ---
 
   /**
--- a/src/contracts/sgho/interfaces/IsGHO.sol
+++ b/src/contracts/sgho/interfaces/IsGHO.sol
@@ -122,7 +122,7 @@ interface IsGHO {
    * @notice Returns the timestamp of the last time the yield index was updated.
    * @return The Unix timestamp of the last update.
    */
-  function lastUpdate() external view returns (uint256);
+  function lastUpdate() external view returns (uint64);
 
   /**
    * @notice Returns the role identifier for the Funds Admin.
--- a/src/contracts/sgho/sGHO.sol
+++ b/src/contracts/sgho/sGHO.sol
@@ -1,0 +1,396 @@
+// SPDX-License-Identifier: agpl-3
+pragma solidity ^0.8.19;
+
+import {ERC4626Upgradeable} from 'openzeppelin-contracts-upgradeable/contracts/token/ERC20/extensions/ERC4626Upgradeable.sol';
+import {ERC20PermitUpgradeable} from 'openzeppelin-contracts-upgradeable/contracts/token/ERC20/extensions/ERC20PermitUpgradeable.sol';
+import {IERC20} from 'openzeppelin-contracts/contracts/token/ERC20/IERC20.sol';
+import {WadRayMath} from 'lib/aave-v3-origin/src/contracts/protocol/libraries/math/WadRayMath.sol';
+import {Initializable} from 'openzeppelin-contracts-upgradeable/contracts/proxy/utils/Initializable.sol';
+import {IAccessControl} from 'openzeppelin-contracts/contracts/access/IAccessControl.sol';
+import {Math} from 'openzeppelin-contracts/contracts/utils/math/Math.sol';
+import {IsGHO} from './interfaces/IsGHO.sol';
+import {ERC20Upgradeable} from 'openzeppelin-contracts-upgradeable/contracts/token/ERC20/ERC20Upgradeable.sol';
+
+/**
+ * @title sGHO Token
+ * @author @kpk
+ * @notice sGHO is an ERC4626 vault that allows users to deposit GHO and earn yield.
+ * @dev This contract implements the ERC4626 standard for tokenized vaults, where the underlying asset is GHO.
+ * It also includes functionalities for yield generation based on a target rate, and administrative roles for managing the contract.
+ */
+contract sGHO is Initializable, ERC4626Upgradeable, ERC20PermitUpgradeable, IsGHO {
+  using WadRayMath for uint256;
+  using Math for uint256;
+
+  address public gho;
+  IAccessControl internal aclManager;
+
+  uint256 public targetRate;
+  uint256 public maxTargetRate;
+  uint256 public supplyCap;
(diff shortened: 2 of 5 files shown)


pavelvm5 (changes requested):

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,421 @@
...
+ * It also includes functionalities for yield generation based on a target rate, and administrative roles for managing the contract.
+ */
+contract sGHO is Initializable, ERC4626Upgradeable, ERC20PermitUpgradeable, IsGHO {
+  using WadRayMath for uint256;
+  using Math for uint256;
+
+  // Storage variables - Optimally packed for gas efficiency
+  uint64 public lastUpdate; // 8 bytes
+  uint16 public targetRate; // 2 bytes
+  address public gho; // 20 bytes
```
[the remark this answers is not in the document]
pavelvm5: Btw, why do we need it at all? We already have `asset()` view function from `ERC4626`.

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,451 @@
...
+
+  /// @custom:storage-location erc7201:gho.storage.sGHO
+  struct sGHOStorage {
+    // Storage variables - Optimally packed for gas efficiency
+    uint64 lastUpdate; // 8 bytes
+    uint16 targetRate; // 2 bytes
+    uint256 supplyCap; // 32 bytes
+    uint256 yieldIndex; // 32 bytes
+    uint256 ratePerSecond; // 32 bytes - cached rate per second for gas efficiency
+  }
```
pavelvm5, line 35:
```suggestion
  struct sGHOStorage {
    // Storage variables - Optimally packed for gas efficiency
    uint168 yieldIndex; // 22 bytes
    uint64 lastUpdate; // 8 bytes
    uint16 targetRate; // 2 bytes
    uint160 supplyCap; // 20 bytes
    uint96 ratePerSecond; // 12 bytes
  }
```
We can pack everything in 2 slots, not 4.

`type(uint96).max` for `ratePerSecond` is 7.92e28, which is more than 249853933300% per year, more than enough.
The same for `type(uint168).max` for `yieldIndex` (3.74e50, more than 76 years of work with 100% APR) and `type(uint160).max` for `supplyCap` (1.46e48 - 1e30 GHO could be deposited).

Luigy-Lemon:
> @Luigy-Lemon
> 
> I left some new comments and also there's one issue, that wasn't resolved:
> 
> * [sGHO - erc4626 vault #5 (comment)](https://github.com/aave-dao/gho-origin/pull/5#discussion_r2236785046) - `SupplyCapExceeded` wasn't deleted, however it's not used.
> 
> Also in Precision.md we would like to see explanation for case, like what happens if updates will happen every second? How much presicion loss will be in such case and so on. Should be very small taking into account 1e27, but for external users this won't be quite easy to understand.

Do you want to know the exact precision loss in respect to compounding because updates are not every second?
or do you mean the precision loss in the yield accrued everytime it gets updated in contract relative to what it would be expected if math was more precise?

pavelvm5:
@Luigy-Lemon 

- Yeah, looks good, but we need to run linter to fix empty lines like here: https://github.com/aave-dao/gho-origin/pull/5/files#diff-a6b551a2e12059c984f17705b48848ce7ecec1a99236117aa4817b92b90e8b2dR55-R58 or extra spaces like here: https://github.com/aave-dao/gho-origin/pull/5/files#diff-a6b551a2e12059c984f17705b48848ce7ecec1a99236117aa4817b92b90e8b2dR35 or split one big line into two or more like here: https://github.com/aave-dao/gho-origin/pull/5/files#diff-a6b551a2e12059c984f17705b48848ce7ecec1a99236117aa4817b92b90e8b2dR43 and here: https://github.com/aave-dao/gho-origin/pull/5/files#diff-a6b551a2e12059c984f17705b48848ce7ecec1a99236117aa4817b92b90e8b2dR25.
I checked it locally, it has found a lot of places to fix.

- I still think, that better to set only executor address in initialize and check if it's zero address:
https://github.com/aave-dao/gho-origin/pull/5/files#diff-a6b551a2e12059c984f17705b48848ce7ecec1a99236117aa4817b92b90e8b2dR77, all other addresses could be added later in the proposal with activation. Also I think executor should have all roles initially, so it could be able to set everything at once.

- And this event: https://github.com/aave-dao/gho-origin/pull/5/files#diff-a6b551a2e12059c984f17705b48848ce7ecec1a99236117aa4817b92b90e8b2dR360. It has invalid description right now, cause it could be updated even if the index is the same. I would split it to two events, one for yield, second for time. Or change description putting the focus on update time first.
I think better to check time, not index, cause if time is already updated index will be the same too:

```
  function _updateYieldIndex() internal {
    sGHOStorage storage $ = _getsGHOStorage();
    
    if ($.lastUpdate != block.timestamp) {  // <-- check time first, not index
      uint176 newYieldIndex = _getCurrentYieldIndex(); // <-- calculate new index
      
      if ($.yieldIndex != newYieldIndex) {
        $.yieldIndex = newYieldIndex; // <-- update new index if it has accrued smth
      }
      
      $.lastUpdate = uint64(block.timestamp); // <-- update time
      emit YieldUpdated(newYieldIndex, block.timestamp); // <-- better to set block.timestamp here, not read from storage
      // I also think that this is a bad name for this event, need to find something more suitable
    }
  }
```

- And this check not needed I guess, don't remembet if I wrote about this or not: https://github.com/aave-dao/gho-origin/pull/5/files#diff-a6b551a2e12059c984f17705b48848ce7ecec1a99236117aa4817b92b90e8b2dR304

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,413 @@
...
+import {IERC4626} from 'openzeppelin-contracts/contracts/interfaces/IERC4626.sol';
+import {IERC20Metadata} from 'openzeppelin-contracts/contracts/token/ERC20/extensions/IERC20Metadata.sol';
+import {Initializable} from 'openzeppelin-contracts-upgradeable/contracts/proxy/utils/Initializable.sol';
+import {Math} from 'openzeppelin-contracts/contracts/utils/math/Math.sol';
+import {SafeCast} from 'openzeppelin-contracts/contracts/utils/math/SafeCast.sol';
+import {IsGHO} from './interfaces/IsGHO.sol';
+import {ERC20Upgradeable} from 'openzeppelin-contracts-upgradeable/contracts/token/ERC20/ERC20Upgradeable.sol';
+import {AccessControlUpgradeable} from 'openzeppelin-contracts-upgradeable/contracts/access/AccessControlUpgradeable.sol';
+import {RescuableACL} from 'lib/aave-v3-origin/lib/solidity-utils/src/contracts/utils/RescuableACL.sol';
+import {RescuableBase, IRescuableBase} from 'lib/aave-v3-origin/lib/solidity-utils/src/contracts/utils/RescuableBase.sol';
```
pavelvm5, line 17: May be sort them with some logic?
Like openzeppelin first, solidity-utils second and interface last or vice versa.

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,413 @@
...
+    return currentAssets >= $.supplyCap ? 0 : $.supplyCap - currentAssets;
+  }
+
+  /**
+   * @inheritdoc IERC4626
+   */
+  function maxMint(address) public view override returns (uint256) {
+    sGHOStorage storage $ = _getsGHOStorage();
+    uint256 currentAssets = totalAssets();
+    return (currentAssets >= $.supplyCap) ? 0 : convertToShares($.supplyCap - currentAssets);
```
pavelvm5, line 143:
```suggestion
    return convertToShares(maxDeposit(receiver));
```

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,413 @@
...
+      )
+    {} catch {}
+
+    // Check actual user balance and adjust assets if necessary
+    // This prevents issues with balance changes during transaction mining
+    // and ensures the transaction doesn't revert due to insufficient balance
+    uint256 actualUserBalance = IERC20(asset()).balanceOf(_msgSender());
+    if (assets > actualUserBalance) {
+      assets = actualUserBalance;
+    }
```
pavelvm5, line 186:
```suggestion
```

I honestly think it's better to revert in this case. Cause if you don't know how much funds you are trying to transfer then something goes completely wrong.

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,413 @@
...
+    // Calculate and cache the new rate per second for gas efficiency
+    if (newRate == 0) {
+      $.ratePerSecond = 0;
+    } else {
+      // Convert targetRate from basis points to ray (1e27 scale)
+      // targetRate is in basis points (e.g., 1000 = 10%)
+      uint256 annualRateRay = uint256(newRate) * RAY / 10000;
+      // Calculate the rate per second (annual rate / seconds in a year)
+      $.ratePerSecond = (annualRateRay / 365 days).toUint96();
+    }
```
pavelvm5, line 258:
```suggestion
    // Convert targetRate from basis points to ray (1e27 scale)
    // targetRate is in basis points (e.g., 1000 = 10%)
    uint256 annualRateRay = uint256(newRate) * RAY / 10000;
    
    // Calculate the rate per second (annual rate / seconds in a year)
    $.ratePerSecond = (annualRateRay / 365 days).toUint96();
```

It will be zero anyway, but this method won't call so frequently to optimize it so hard, so less code is better here, I guess.

src/contracts/sgho/sGHO.sol
Luigy-Lemon: I followed the function in the StataToken currently in production. I can remove this condition but should i use a custom error here or maintain the default from SafeERC20?

pavelvm5: Hi, the reason we use this logic inside `StataToken` is that we use rebase balances for `aTokens`, so we are not able to calculate exact input amount during `depositPermit` call. Cause it depends on a lot of factors.

But in case of `Gho` it's not needed, cause `Gho` itself is non-rebasable. So better to remove it.
I guess we don't need to introduce any type of check here, if allowance is not given for some reason, then deposit will revert with default error.

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,392 @@
...
+   * @param _gho       Address of the underlying GHO token.
+   * @param _supplyCap The total supply cap for the vault.
+   * @param _executor  The address that will be granted the DEFAULT_ADMIN_ROLE.
+   */
+  function initialize(
+    address _gho,
+    uint160 _supplyCap,
+    address _executor
+  ) public payable initializer {
+    if (_gho == address(0)) revert ZeroAddressNotAllowed();
```
pavelvm5, line 83:
```suggestion
  /**
   * @notice Initializer for the sGHO vault.
   * @param gho_       Address of the underlying GHO token.
   * @param supplyCap_ The total supply cap for the vault.
   * @param executor_  The address that will be granted the DEFAULT_ADMIN_ROLE.
   */
  function initialize(
    address gho_,
    uint160 supplyCap_,
    address executor_
  ) public payable initializer {
    if (gho_ == address(0) || executor_ == address(0)) revert ZeroAddressNotAllowed();
```

`_name` for internal variables/functions, `name_` for external arguments, accoarding to solidity docs.
Also zero address check was skipped.

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,392 @@
...
+    __ERC20Permit_init('sGHO');
+    __AccessControl_init();
+
+    _grantRole(DEFAULT_ADMIN_ROLE, _executor);
+
+    sGHOStorage storage $ = _getsGHOStorage();
+    $.supplyCap = _supplyCap;
+    $.yieldIndex = RAY;
+    $.lastUpdate = uint64(block.timestamp);
+    $.ratePerSecond = 0; // Initial rate is 0, so ratePerSecond is 0 (no yield initially)
```
pavelvm5, line 96:
```suggestion
    // Initial ratePerSecond and targetRate is 0 (no yield initially)
```

We need to set both of them to zero, or skip set for both. Don't understand why rate was set to zero and targetRate not.

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,392 @@
...
+    uint256 ghoBalance = IERC20(asset()).balanceOf(address(this));
+    uint256 maxRedeemShares = super.maxRedeem(owner);
+    uint256 sharesForBalance = convertToShares(ghoBalance);
+    return maxRedeemShares < sharesForBalance ? maxRedeemShares : sharesForBalance;
+  }
+
+  /**
+   * @inheritdoc IERC4626
+   */
+  function maxDeposit(address receiver) public view override returns (uint256) {
```
pavelvm5, line 131:
```suggestion
  function maxDeposit(address) public view override returns (uint256) {
```

Solidity warning.

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,392 @@
...
+  ) public override returns (uint256) {
+    _updateYieldIndex();
+    return super.redeem(shares, receiver, owner);
+  }
+
+  /**
+   * @notice Returns the total amount of GHO managed by the vault.
+   * @dev This is calculated based on the total supply of sGHO and the current yield index.
+   * @return The total amount of GHO assets.
+   */
```
pavelvm5, line 215:
```suggestion
  /**
   * @inheritdoc IERC4626
   */
```

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,392 @@
...
+   * @notice Updates the yield index to accrue yield up to the current timestamp.
+   * @dev This function modifies state and is called before any operation that depends on the yield index.
+   * Uses SafeCast to prevent overflow when casting to uint176. If overflow occurs, the transaction will revert
+   * instead of silently wrapping, protecting user rewards.
+   */
+  function _updateYieldIndex() internal {
+    sGHOStorage storage $ = _getsGHOStorage();
+    if ($.lastUpdate != block.timestamp) {
+      uint256 newYieldIndex = _getCurrentYieldIndex();
+      $.yieldIndex = newYieldIndex.toUint176();
```
pavelvm5, line 336:
```suggestion
      uint176 newYieldIndex = _getCurrentYieldIndex();
      $.yieldIndex = newYieldIndex;
```

Already `safeCast`ed.

pavelvm5 (commented):
I think these are the last comments
Fix them carefully please, cause some comments were repeated several times.

src/contracts/sgho/precision_analysis/PRECISION.md
```diff
@@ -0,0 +1,290 @@
...
+---
+
+## Yield Accrual and Indexing
+
+- Yield is accrued linearly between state updates (deposit, withdraw, mint, redeem), but compounds across multiple updates.
+- The yield index (`yieldIndex`) is always stored and updated in **ray** (1e27).
+- The target rate (`targetRate`) is set in **basis points** (1e4 = 100%).
+- The rate per second (`ratePerSecond`) is cached for gas efficiency.
+
+> **Note:** Even if the target rate is set to the maximum (50% APR) and the yield is compounded daily due to user actions for 100 years, the `yieldIndex` will not exceed approximately `5e29`. This demonstrates that the system is robust against overflow and extreme long-term compounding scenarios.
```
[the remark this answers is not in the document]
Luigy-Lemon: Yeah. This is wrong, I did the math when I was storing it in 8 decimals. It would be around 5e47 compounded daily.

src/contracts/sgho/interfaces/IsGHO.sol
```diff
@@ -0,0 +1,154 @@
...
+   * @notice Returns the current target annual percentage rate (APR) for yield generation.
+   * @dev The rate is expressed in basis points (1% = 100).
+   * @return The target rate in basis points.
+   */
+  function targetRate() external view returns (uint16);
+
+  /**
+   * @notice Returns the current rate per second for yield generation.
+   * @dev The rate is expressed in basis points (1% = 100).
+   * @return The rate per second in basis points multiplied by 10^18.
```
pavelvm5, line 96:
```suggestion
   * @return The rate per second in basis points multiplied by 10^27.
```

Ray, not Wad, sorry, my fault.

Everything else looks fine, thanks

CheyenneAtapour (approved):

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+  uint176 private constant RAY = 1e27;
+
+  /// @custom:storage-location erc7201:gho.storage.sGHO
+  struct sGHOStorage {
+    // Storage variables - Optimally packed for gas efficiency
+    uint176 yieldIndex; // 22 bytes - current yield index for share/asset conversion
+    uint64 lastUpdate; // 8 bytes - timestamp of last yield index update
+    uint16 targetRate; // 2 bytes - target annual yield rate in basis points (e.g., 1000 = 10%)
+    uint160 supplyCap; // 20 bytes - maximum total assets allowed in the vault
+    uint96 ratePerSecond; // 12 bytes - cached rate per second for gas efficiency
```
miguelmtzinf: it's not in RAY terms?
2**96/1e27 = ~79 ?

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+    return (($.yieldIndex * growthFactor) / RAY).toUint176();
+  }
+
+  /**
+   * @notice Updates the yield index to accrue yield up to the current timestamp.
+   * @dev This function modifies state and is called before any operation that depends on the yield index.
+   * Uses SafeCast to prevent overflow when casting to uint176. If overflow occurs, the transaction will revert
+   * instead of silently wrapping, protecting user rewards.
+   */
+  function _updateYieldIndex() internal {
```
miguelmtzinf: please follow Solidity style guidelines: https://docs.soliditylang.org/en/latest/style-guide.html#order-of-layout

> Contract elements should be laid out in the following order:
> Pragma statements
> Import statements
> Events
> Errors
> Interfaces
> Libraries
> Contracts
> Inside each contract, library or interface, use the following order:
> Type declarations
> State variables
> Events
> Errors
> Modifiers
> Functions
>

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+  bytes32 private constant sGHOStorageLocation =
+    0xfdf74a24098989caa4d9d232df283137a30d85fb47ad37b31478f919573b9800;
+
+  function _getsGHOStorage() private pure returns (sGHOStorage storage $) {
+    assembly {
+      $.slot := sGHOStorageLocation
+    }
+  }
+
+  // Constants (stored in bytecode, not storage)
```
miguelmtzinf: i d rm this, and provide proper natspec docs to these constants

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+   */
+  constructor() {
+    _disableInitializers();
+  }
+
+  /**
+   * @notice Initializer for the sGHO vault.
+   * @param gho_       Address of the underlying GHO token.
+   * @param supplyCap_ The total supply cap for the vault.
+   * @param executor_  The address that will be granted the DEFAULT_ADMIN_ROLE.
```
miguelmtzinf:
```suggestion
   * @param gho The address of the underlying GHO token.
   * @param supplyCap The total supply cap for the vault.
   * @param executor The address that will be granted the `DEFAULT_ADMIN_ROLE`.
```

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+    assembly {
+      $.slot := sGHOStorageLocation
+    }
+  }
+
+  // Constants (stored in bytecode, not storage)
+  uint16 public constant MAX_SAFE_RATE = 5000; // Maximum safe annual yield rate in basis points (50%)
+  bytes32 public constant FUNDS_ADMIN_ROLE = 'FUNDS_ADMIN'; // Role for managing rescued funds
+  bytes32 public constant YIELD_MANAGER_ROLE = 'YIELD_MANAGER'; // Role for managing yield rates and supply caps
+  bytes32 public constant PAUSE_GUARDIAN_ROLE = 'PAUSE_GUARDIAN_ROLE'; // Role for managing pause and unpause
```
miguelmtzinf:
```suggestion
  bytes32 public constant PAUSE_GUARDIAN_ROLE = 'PAUSE_GUARDIAN'; // Role for managing pause and unpause
```

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+  // RAY is used for high-precision mathematical operations to avoid rounding errors
+  uint176 private constant RAY = 1e27;
+
+  /// @custom:storage-location erc7201:gho.storage.sGHO
+  struct sGHOStorage {
+    // Storage variables - Optimally packed for gas efficiency
+    uint176 yieldIndex; // 22 bytes - current yield index for share/asset conversion
+    uint64 lastUpdate; // 8 bytes - timestamp of last yield index update
+    uint16 targetRate; // 2 bytes - target annual yield rate in basis points (e.g., 1000 = 10%)
+    uint160 supplyCap; // 20 bytes - maximum total assets allowed in the vault
```
miguelmtzinf: is this in asset terms or whole units?

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+        sig.s
+      )
+    {} catch {}
+
+    return deposit(assets, receiver);
+  }
+
+  /**
+   * @inheritdoc IERC4626
+   */
```
miguelmtzinf: I d suggest to do this in general, but in thsi fn in particular I think it make sense to write its own docs as it's doing something different to 4626
```suggestion
  /// @inheritdoc IERC4626
```

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+    _updateYieldIndex();
+    super._update(from, to, value);
+  }
+
+  /**
+   * @notice Override _checkRescueGuardian to check for FUNDS_ADMIN role
+   * @dev This function reverts if the caller doesn't have the FUNDS_ADMIN role
+   */
+  function _checkRescueGuardian() internal view override {
+    if (!hasRole(FUNDS_ADMIN_ROLE, _msgSender())) {
```
miguelmtzinf: `FUNDS_ADMIN_ROLE` does not seem appropiate for the rescue role imho. The GSM uses `TOKEN_RESCUER_ROLE`, for reference.

src/contracts/sgho/sGHO.sol
```diff
@@ -0,0 +1,380 @@
...
+   * @inheritdoc IsGHO
+   */
+  function setTargetRate(uint16 newRate) public onlyRole(YIELD_MANAGER_ROLE) {
+    sGHOStorage storage $ = _getsGHOStorage();
+    // Update the yield index before changing the rate to ensure proper accrual
+    if (newRate > MAX_SAFE_RATE) {
+      revert RateMustBeLessThanMaxRate();
+    }
+    _updateYieldIndex();
+    $.targetRate = newRate;
```
miguelmtzinf: why do we need to store the `targetRate` if `ratePerSecond` is the only thing that it's used for index updates?

src/contracts/sgho/sGHO.sol
efecarranza: Expected lifetime of this version of sGho is 1-2 years, should be ok.

src/contracts/sgho/sGHO.sol
efecarranza: because this is used elsewhere with total assets, i'm assuming it's in asset terms

PR #6, opened 2025-08-11
Feat: add sGhoSteward

pavelvm5:
Adding code for `sGhoSteward` that can update `supplyCap` and `targetRate` in `sGho`.
For this to work, this steward should be granted with `YIELD_MANAGER_ROLE`.

For convenience, the `targetRate` is divided into three components and a special role is allocated for updating each variable.


kyzia551 (approved):

tests/misc/sGhoSteward.t.sol
```diff
@@ -0,0 +1,301 @@
...
+    vm.stopPrank();
+    vm.startPrank(ghoCommittee);
+
+    newConfig = IsGhoSteward.RateConfig({
+      amplification: 200_00, // 3 * AMPLIFICATION_DENOMINATOR
+      floatRate: 200, // 2%
+      fixedRate: 300 // 3%
+    });
+
+    vm.expectRevert();
```
CheyenneAtapour: Perhaps explicitly verify that the revert reason is because of the role?

src/contracts/misc/sGhoSteward.sol
```diff
@@ -0,0 +1,128 @@
...
+ * @author BGD Labs
+ * @notice Helper contract for managing rate and supply cap parameters for sGho.
+ */
+contract sGhoSteward is AccessControl, IsGhoSteward {
+  using SafeCast for uint256;
+
+  /// @notice Current rate parameters
+  RateConfig public rateConfig;
+
+  IsGHO public immutable sGHO;
```
efecarranza: In the other contracts in the repo these are prefixed and then params on functions don't have prefix/suffix from what i've seen

src/contracts/misc/sGhoSteward.md
```diff
@@ -0,0 +1,126 @@
...
+
+- Allows updating of `targetRate` and `supplyCap` of `sGHO` through this contract.
+- Implements the `targetRate` formula as a composition of three parameters:
+
+  ```targetRate = AmplificationFactor * FloatRate + FixedRate```
+
+- Supports updating with each rate parameter *individually* or *multiple parameters simultaneously*.
+- Integrates **Role-Based Access Control** from OpenZeppelin, enabling secure assignment, modification, and revocation of roles.
+- Provides functions to view the current configuration and pre-calculate the `targetRate` for any given configuration.
+- Enforces a hard cap of `50%` for the computed `targetRate`, regardless of configuration input.
```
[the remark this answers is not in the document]
pavelvm5: Description has been fixed.

src/contracts/misc/sGhoSteward.sol
```diff
@@ -0,0 +1,128 @@
...
+  uint16 public immutable MAX_RATE;
+
+  uint16 public constant AMPLIFICATION_DENOMINATOR = 100_00;
+
+  bytes32 public constant AMPLIFICATION_MANAGER_ROLE = keccak256('AMPLIFICATION_MANAGER_ROLE');
```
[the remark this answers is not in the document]
pavelvm5: Added.

efecarranza:
@CheyenneAtapour updated! thanks!

src/contracts/misc/sGhoSteward.sol
```diff
@@ -0,0 +1,148 @@
...
+
+    if (rateConfigCopy.fixedRate != rateConfig.fixedRate) {
+      _checkRole(FIXED_RATE_MANAGER_ROLE);
+
+      isRateChanged = true;
+      rateConfigCopy.fixedRate = rateConfig.fixedRate;
+    }
+
+    if (!isRateChanged) {
+      revert SameValue();
```
Kogaroshi: maybe a more explicit error, such as `RateNotChanged` ?

src/contracts/misc/sGhoSteward.sol
```diff
@@ -0,0 +1,148 @@
...
+ * @notice Helper contract for managing rate and supply cap parameters for sGho.
+ */
+contract sGhoSteward is AccessControl, IsGhoSteward {
+  using SafeCast for uint256;
+
+  /// @notice Current rate parameters
+  RateConfig internal _rateConfig;
+
+  /// @notice sGho contract address
+  IsGHO internal immutable _sGHO;
```
Kogaroshi: rename all `sGHO` to `sGho` like in #22

src/contracts/misc/interfaces/IsGhoSteward.sol
```diff
@@ -0,0 +1,135 @@
...
+   *       where `AMPLIFICATION_DENOMINATOR` is 100_00,
+   *       `amplification`, `floatRate` and `fixedRate` are `uint16`
+   */
+  struct RateConfig {
+    /// @notice Amplification factor
+    uint16 amplification;
+    /// @notice Representative of market conditions
+    uint16 floatRate;
+    /// @notice Nominal Amount
+    uint16 fixedRate;
```
Kogaroshi, line 19: we might prefer having those rates as `uint32` : there is enough space in the struct, and `uint16` max value is 65,535, so 650% rate, tho it seems improbable, better be safe there.

tests/misc/sGhoSteward.t.sol
```diff
@@ -0,0 +1,384 @@
...
+
+    assertTrue(steward.hasRole(AMPLIFICATION_MANAGER_ROLE, ghoCommittee));
+    assertTrue(steward.hasRole(FLOAT_RATE_MANAGER_ROLE, ghoCommittee));
+    assertTrue(steward.hasRole(FIXED_RATE_MANAGER_ROLE, ghoCommittee));
+    assertTrue(steward.hasRole(SUPPLY_CAP_MANAGER_ROLE, ghoCommittee));
+
+    assertEq(address(steward.sGHO()), address(sGho));
+  }
+
+  function test_setRateConfig() public {
```
Kogaroshi, line 58: please add a fuzz test with sanitized `IsGhoSteward.RateConfig` input.

src/contracts/misc/interfaces/IsGhoSteward.sol
pavelvm5: This makes no sense, since the final rate is stored in a `uint16`. 

The only two values that can truly be higher than `uint16` is `amplification` or `floatRate`, and only if `floatRate` or `amplification` is set to the minimum possible non-zero value. However, in that case, increasing `floatRate` and `amplification` is required from both roles at the same time with cooperation and not just in 1 call from any address. Everything else is pointless, due to the point described above.

Otherwise, the point of having three different roles is lost, since any of the three roles can effectively set any possible rate, up to the maximum. However, it's clear that `floatRate` or `amplification` have less impact on the system than `fixedRate`. Therefore, these roles should be installed with different security priorities.

On the contrary, having `uint16` variables is much safer, because if one of the roles is compromised, it will not be able to make the minimum possible `floatRate` lead to the maximum possible `targetRate`, and also the same with `amplification`.

src/contracts/misc/sGhoSteward.sol
pavelvm5: Fixed.

src/contracts/misc/sGhoSteward.sol
pavelvm5: Fixed.

tests/misc/sGhoSteward.t.sol
pavelvm5: Added.

pavelvm5:
Everything is fixed or answered. Since I don't have write access you'll need to change base branch from `main` to `feat/sGho`.  Because everything was originally created to the `main` branch the diff is huge.

@efecarranza, @Kogaroshi

src/contracts/misc/sGhoSteward.sol
```diff
@@ -0,0 +1,148 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import {AccessControl} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/access/AccessControl.sol';
+import {SafeCast} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/utils/math/SafeCast.sol';
+
+import {IsGhoSteward} from './interfaces/IsGhoSteward.sol';
+import {IsGho} from '../sgho/interfaces/IsGHO.sol';
```
Kogaroshi: pls also rename the file for consistency

Kogaroshi:
> Everything is fixed or answered. Since I don't have write access you'll need to change base branch from `main` to `feat/sGho`. Because everything was originally created to the `main` branch the diff is huge.
> 
> @efecarranza, @Kogaroshi

Done, pls review that no wanted changes got mixed in this PR (it seems there is changes to `GhoDirectFacilitator` in here as well now)

src/contracts/misc/sGhoSteward.sol
pavelvm5: image

Guys, there's no consistency in sGHO name, cause in whole repo it's named in Camel-case.
I guess files should be named SGho, SGhoSteward to make everything consistence.

Kogaroshi: Yes, that's what I meant, to rename files `sGho` & `IsGho`.
Rest of the files we write `Gho` so we should keep the same here

Kogaroshi: or `SGho` & `ISGho`, but then we need to rename contract name and test files to match.

pavelvm5: Done

Kogaroshi (approved):
Lint CI fails, otherwise lgtm

src/contracts/misc/interfaces/IsGhoSteward.sol
CheyenneAtapour, line 1: I will add some nits, but these are nbd and can be ignored

src/contracts/misc/interfaces/IsGhoSteward.sol
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * @notice Event is emitted whenever the `rateConfig` is updated.
+   * @param caller Message sender, who initiated the update
+   * @param targetRate Target rate installed in `sGHO` after update
```
CheyenneAtapour: nit: I don't like this terminology of `installed` used throughout this contract, but rather perhaps a word like `set`

src/contracts/misc/interfaces/IsGhoSteward.sol
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * @dev Attempted to set zero address.
+   */
+  error ZeroAddress();
+
+  /**
+   * @dev Attempted to set rate greater than `MAX_RATE` defined in `sGHO`.
+   */
+  error RateTooBig();
```
CheyenneAtapour: nit: For me, `MaxRateExceeded` or similar would be better and more formal

src/contracts/misc/interfaces/IsGhoSteward.sol
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * @dev Attempted to set the same rate, which is already installed.
+   */
+  error SameRate();
```
CheyenneAtapour: nit: Perhaps `RateUnchanged`

src/contracts/misc/interfaces/IsGhoSteward.sol
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * @dev Attempted to set the same supplyCap, which is already installed.
+   */
+  error SameSupplyCap();
```
CheyenneAtapour: nit: `SupplyCapUnchanged`

src/contracts/misc/interfaces/IsGhoSteward.sol
```diff
@@ -0,0 +1,140 @@
...
+   *   - `FIXED_RATE_MANAGER_ROLE`
+   *
+   * @param rateConfig_ Set of parameters for calculating `targetRate`
+   * @return targetRate `targetRate` set in `sGHO`
+   */
+  function setRateConfig(RateConfig calldata rateConfig_) external returns (uint16);
+
+  /**
+   * @notice Updates `supplyCap` on `sGHO`.
+   * @dev Could be updated to any `uint160` value.
```
CheyenneAtapour: nit: Is it worth it so say here "reverts otherwise". I see the safecast is what enforces this. For example in natspec of `previewTargetRate` it is mentioned that the contract reverts if the new `targetRate` value is out of bounds

src/contracts/misc/interfaces/IsGhoSteward.sol
```diff
@@ -0,0 +1,140 @@
...
+   */
+  function previewTargetRate(RateConfig calldata rateConfig_) external view returns (uint16);
+
+  /**
+   * @notice Returns current `rateConfig`.
+   */
+  function getRateConfig() external view returns (RateConfig memory);
+
+  /**
+   * @notice Returns `sGHO` address.
```
CheyenneAtapour: nit: Maybe confusing slightly due to the inaccuracy because it technically returns an interface, not address type

src/contracts/misc/sGhoSteward.md
```diff
@@ -0,0 +1,126 @@
...
+
+**sGhoSteward** is a steward contract designed to manage and update configuration parameters of `sGHO`.
+It provides role-based access-controlled mechanisms to safely adjust the `targetRate` and `supplyCap` for the Savings GHO Vault.
+
+## Key features
+
+- Allows updating of `targetRate` and `supplyCap` of `sGHO` through this contract.
+- Implements the `targetRate` formula as a composition of three parameters:
+
+  `targetRate = AmplificationFactor * FloatRate + FixedRate`
```
CheyenneAtapour: nit: is it bad that this differs from the description within `IsGhoSteward` which includes the division by the `AMPLIFICATION_DENOMINATOR`?

src/contracts/misc/sGhoSteward.md
```diff
@@ -0,0 +1,126 @@
...
+
+- Supports updating with each rate parameter _individually_ or _multiple parameters simultaneously_.
+- Integrates **Role-Based Access Control** from OpenZeppelin, enabling secure assignment, modification, and revocation of roles.
+- Provides functions to view the current configuration and pre-calculate the `targetRate` for any given configuration.
+- Enforces a hard cap of `50%` for the computed `targetRate`, reverting the transaction if the changes do not meet the condition.
```
CheyenneAtapour: This is a bit inaccurate. The steward fetches this rate from the implementation of `sGho.sol`. If a different implementation is used, it could be different than 50%. The steward contract itself only enforces the MAX_SAFE_RATE it fetches from the implementation of `sGho.sol`

src/contracts/misc/sGhoSteward.md
```diff
@@ -0,0 +1,126 @@
...
+
+Current configuration:
+
+```solidity
+amplification: 50_00   // 0.5
+floatRate:     4_00    // 4%
+fixedRate:     2_00    // 2%
+```
+
+Current `targetRate` is calculated as `50_00 * 4_00 / 100_00 + 2_00 = 4_00`. `4_00` refers to 4%.
```
CheyenneAtapour, line 63: The example here also has the division by the `AMPLIFICATION_FACTOR`, so perhaps is helpful to write it like that always to avoid confusion

src/contracts/misc/sGhoSteward.md
```diff
@@ -0,0 +1,126 @@
...
+`10_000` is `AMPLIFICATION_FACTOR`, which is constant. Every parameter has `uint16` type with max value up to `65_535`.
```
CheyenneAtapour: should this be written as `100_00` as above?
```suggestion
`100_00` is `AMPLIFICATION_FACTOR`, which is constant. Every parameter has `uint16` type with max value up to `65_535`.
```

src/contracts/misc/sGhoSteward.md
```diff
@@ -0,0 +1,126 @@
...
+- Access is restricted to addresses with the `SUPPLY_CAP_MANAGER_ROLE`.
+- The maximum cap allowed within sGHO is `uint160` (less than the `uint256` input type of this function). Any attempt to set a value above this threshold results in a revert.
+- Similarly, attempting to set the same value as the current `supplyCap` will also cause the transaction to revert.
+
+## Contract Summary
+
+| Function                               | Description                                        | Required Role               |
+| :------------------------------------- | :------------------------------------------------- | :-------------------------- |
+| `setRateConfig(RateConfig newConfig)`  | Updates amplification, float, and fixed rates      | Corresponding Manager Roles |
+| `setSupplyCap(uint160 newCap)`         | Updates the maximum allowed sGHO supply            | `SUPPLY_CAP_MANAGER_ROLE`   |
```
CheyenneAtapour: Should it be the correct function signature here? It's mentioned above how the input type is `uint256` already, so I think using the wrong function signature here can be confusing, and instead should match the contract.
```suggestion
| `setSupplyCap(uint256 supplyCap)`         | Updates the maximum allowed sGHO supply            | `SUPPLY_CAP_MANAGER_ROLE`   |
```
The parameter is also incorrect, but I don't think that matters as much

src/contracts/misc/sGhoSteward.md
```diff
@@ -0,0 +1,126 @@
...
+| `getRateConfig()`                      | Returns the current configuration                  | Public                      |
```
CheyenneAtapour:
```suggestion
| `getRateConfig()`                      | Returns the current rate configuration                  | Public                      |
```

nit

tests/misc/sGhoSteward.t.sol
```diff
@@ -0,0 +1,412 @@
...
+    IsGhoSteward.RateConfig memory configAfterUpdate = steward.getRateConfig();
+
+    assertEq(configAfterUpdate.amplification, 100_00);
+    assertEq(configAfterUpdate.floatRate, 2_00);
+    assertEq(configAfterUpdate.fixedRate, 2_00);
+
+    vm.startPrank(executor);
+
+    steward.revokeRole(FIXED_RATE_MANAGER_ROLE, ghoCommittee);
+    steward.revokeRole(FLOAT_RATE_MANAGER_ROLE, ghoCommittee);
```
CheyenneAtapour, line 109: Would it also be good to show the ghoCommittee does not have `DEFAULT_ADMIN_ROLE` here or within `test_initial`? 

Maybe implicitly done with revoking the required role and showing it reverts

tests/misc/sGhoSteward.t.sol
```diff
@@ -0,0 +1,412 @@
...
+    vm.startPrank(ghoCommittee);
+
+    steward.setSupplyCap(type(uint160).max);
+
+    uint256 supplyCapAfterUpdate = sgho.supplyCap();
+    assertEq(supplyCapAfterUpdate, type(uint160).max);
+
+    vm.expectRevert(abi.encodeWithSelector(IsGhoSteward.SameSupplyCap.selector));
+    steward.setSupplyCap(type(uint160).max);
+  }
```
CheyenneAtapour, line 326: How about a test showing trying to set supply cap to a value greater than `type(uint160).max` reverts?

src/contracts/misc/interfaces/IsGhoSteward.sol
pavelvm5: Everything was fixed.

Kogaroshi (dismissed):
Please fix CI & lint, otherwise lgtm

Kogaroshi (approved):

PR #8, opened 2025-08-27
fix : modify GSM signature encoding

Kogaroshi:
Modify the logic to encode and hash signatures in the GSM to respect EIP712, and updated the related unit tests with the new foundry cheatcodes.


Kogaroshi:
Fixes #10

tests/unit/TestGhoBase.t.sol
```diff
@@ -722,4 +724,30 @@ contract TestGhoBase is Test, Constants, Events {
     );
     (v, r, s) = vm.sign(ownerPk, outerHash);
   }
+
+  function _getBuyTypedDataHash(
```
miguelmtzinf, line 728:
```suggestion
  function _getBuyAssetTypedDataHash(
```
just for clarity

tests/unit/TestGhoBase.t.sol
```diff
@@ -722,4 +724,30 @@ contract TestGhoBase is Test, Constants, Events {
...
+      keccak256(
+        abi.encodePacked(
+          '\x19\x01',
+          GHO_GSM.DOMAIN_SEPARATOR(),
+          vm.eip712HashStruct('BuyAssetWithSig', abi.encode(params))
+        )
+      );
+  }
+
+  function _getSellTypedDataHash(
```
miguelmtzinf, line 741:
```suggestion
  function _getSellAssetTypedDataHash(
```

tests/mocks/JsonBindings.sol
miguelmtzinf, line 1: should we git-ignore this file? as it's automatically generated

Kogaroshi: it was generated using `forge bind-json`, it is done once and not needed to be automatic or re-created as long as new types are not introduced or the current ones modified

PR #20, opened 2025-12-29
sGho Updates

efecarranza:
Builds on top of: #5 sGho
I'd make both of these PRs target an sgho side-branch rather than main, merge that one first, then see the changes here before going to main. Have a quick look back by Certora potentially.

Separates tests into different files to make more manageable.
Addresses some style issues.


efecarranza:
@miguelmtzinf I rebased with feat/sgho but I changed the names from sGHO to sGho and it won't show the changes. tried readding sGHO but it won't add changes to git. any ideas? 😬 tried debugging for a while

PR #21, opened 2026-01-06
fix: Gsm initialize

CheyenneAtapour:
This PR adds the `initializer` modifier to the constructor of the Gsm contract so that the implementation contract of the Gsm is already initialized upon deployment. This will make deployments of the Gsm contract easier and safer.

### DIFF
--- a/src/contracts/facilitators/gsm/Gsm.sol
+++ b/src/contracts/facilitators/gsm/Gsm.sol
@@ -92,7 +92,11 @@ contract Gsm is AccessControl, VersionedInitializable, EIP712, IGsm {
    * @param underlyingAsset The address of the collateral asset
    * @param priceStrategy The address of the price strategy
    */
-  constructor(address ghoToken, address underlyingAsset, address priceStrategy) EIP712('GSM', '1') initializer {
+  constructor(
+    address ghoToken,
+    address underlyingAsset,
+    address priceStrategy
+  ) EIP712('GSM', '1') initializer {
     require(ghoToken != address(0), 'ZERO_ADDRESS_NOT_VALID');
     require(underlyingAsset != address(0), 'ZERO_ADDRESS_NOT_VALID');
     require(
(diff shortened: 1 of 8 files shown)


efecarranza (approved):

tests/unit/TestGhoBase.t.sol
avniculae, line 277: I'd call this ghoGsm4626 and define it here, to align to `Gsm gsm = ` from above

tests/unit/TestGsm.t.sol
```diff
@@ -50,6 +50,7 @@ contract TestGsm is TestGhoBase {
       address(USDX_TOKEN),
       address(GHO_GSM_FIXED_PRICE_STRATEGY)
     );
+    gsm = Gsm(address(new AdminUpgradeabilityProxy(address(gsm), SHORT_EXECUTOR, '')));
```
avniculae, line 53: I think we should create a helper function to deploy gsm impl, proxy and return casted proxy. We can use it in all tests and in TestGhoBase as well

tests/unit/TestGsm4626.t.sol
```diff
@@ -46,6 +46,7 @@ contract TestGsm4626 is TestGhoBase {
       address(USDX_4626_TOKEN),
       address(GHO_GSM_4626_FIXED_PRICE_STRATEGY)
     );
+    gsm = Gsm4626(address(new AdminUpgradeabilityProxy(address(gsm), SHORT_EXECUTOR, '')));
```
avniculae, line 49: similar for gsm4626 as here https://github.com/aave-dao/gho-origin/pull/21/changes#r2863675253

src/contracts/facilitators/gsm/Gsm.sol
avniculae, line 1: I see no test that checks implementation is initialized after deployment. Can you pls point me to it? (or add it if we don't have one)

we should have one for Gsm, and one for Gsm4626

avniculae: also, should we add `initializer` to GhoReserve as well?

avniculae:
This is not needed anymore since gho-origin updated aave-v3-origin to v3.6, which has a new version of VersionedInitializable (introduced in v3.4) that blocks initialize in constructor.

PR #24, opened 2026-01-08
fix: Use initializable over versionedInitializable

CheyenneAtapour:
Changes `GhoReserve`, `FixedFeeStrategyFactory` and `Gsm` contracts to use `Initializable` instead of `VersionedInitializable` to help ensure implementation contracts are not left uninitialized


lib/aave-v3-origin
miguelmtzinf, line 1: don't change

src/contracts/facilitators/gsm/feeStrategy/FixedFeeStrategyFactory.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.10;
 
 import {EnumerableSet} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/utils/structs/EnumerableSet.sol';
-import {VersionedInitializable} from 'aave-v3-origin/contracts/misc/aave-upgradeability/VersionedInitializable.sol';
+import {Initializable} from 'src/contracts/dependencies/solidity-utils/src/contracts/transparent-proxy/Initializable.sol';
```
miguelmtzinf: this should use OZ. path: aave-v3-origin/solidity-utils/openzeppeling
makes sense to add to remappings

lib/aave-v3-origin
CheyenneAtapour: Done

src/contracts/facilitators/gsm/feeStrategy/FixedFeeStrategyFactory.sol
CheyenneAtapour: Done

src/contracts/facilitators/gsm/feeStrategy/FixedFeeStrategyFactory.sol
```diff
@@ -77,9 +77,4 @@ contract FixedFeeStrategyFactory is VersionedInitializable, IFixedFeeStrategyFac
   function REVISION() public pure virtual override returns (uint256) {
```
miguelmtzinf: let's i) use a constant instead ii) use the modifier `reinitializer(REVISION)` and iii) call `_disableInitializers` in constructor.
reference: https://github.com/aave/aave-v4/blob/main/src/spoke/instances/SpokeInstance.sol

src/contracts/facilitators/gsm/GhoReserve.sol
```diff
@@ -14,7 +14,7 @@ import {IGhoReserve} from 'src/contracts/facilitators/gsm/interfaces/IGhoReserve
  * @notice It allows approved entities to withdraw and return GHO funds, with a defined maximum withdrawal capacity per entity.
  * @dev To be covered by a proxy contract.
  */
-contract GhoReserve is AccessControl, VersionedInitializable, IGhoReserve {
+contract GhoReserve is AccessControl, Initializable, IGhoReserve {
```
miguelmtzinf:
```suggestion
contract GhoReserve is Initializable, AccessControl, IGhoReserve {
```

src/contracts/facilitators/gsm/Gsm.sol
```diff
@@ -21,7 +21,7 @@ import {IGsm} from 'src/contracts/facilitators/gsm/interfaces/IGsm.sol';
  * @notice GHO Stability Module. It provides buy/sell facilities to go to/from an underlying asset to/from GHO.
  * @dev To be covered by a proxy contract.
  */
-contract Gsm is AccessControl, VersionedInitializable, EIP712, IGsm {
+contract Gsm is AccessControl, Initializable, EIP712, IGsm {
```
miguelmtzinf:
```suggestion
contract Gsm is Initializable, EIP712, AccessControl, IGsm {
```
because EIP712 has no storage

src/contracts/facilitators/gsm/Gsm.sol
```diff
@@ -102,6 +102,7 @@ contract Gsm is AccessControl, VersionedInitializable, EIP712, IGsm {
     GHO_TOKEN = ghoToken;
     UNDERLYING_ASSET = underlyingAsset;
     PRICE_STRATEGY = priceStrategy;
+    _disableInitializers();
```
miguelmtzinf: let's do it at the beginning

CheyenneAtapour: Done

src/contracts/facilitators/gsm/Gsm.sol
CheyenneAtapour: Done

src/contracts/facilitators/gsm/GhoReserve.sol
CheyenneAtapour: Done

src/contracts/facilitators/gsm/feeStrategy/FixedFeeStrategyFactory.sol
CheyenneAtapour: Done

src/contracts/facilitators/gsm/GhoReserve.sol
avniculae, line 49: why don't we call `_disableInitializers` for GhoReserve?

src/contracts/facilitators/gsm/feeStrategy/FixedFeeStrategyFactory.sol
```diff
@@ -24,7 +30,7 @@ contract FixedFeeStrategyFactory is VersionedInitializable, IFixedFeeStrategyFac
    * @param feeStrategiesList List of fee strategies
    * @dev Assumes that the addresses provided are deployed FixedFeeStrategy contracts
    */
-  function initialize(address[] memory feeStrategiesList) external initializer {
+  function initialize(address[] memory feeStrategiesList) external reinitializer(REVISION) {
```
avniculae, line 34: I'd still use `initializer` given that we don't upgrade proxies, but rather deploy new ones, wdyt? having the possibility of upgrading the proxy with the current impl can be error-prone

avniculae: if we do this, I'd revert REVISION to uint256 for backwards compatibility, since it will not be used anyways

src/contracts/facilitators/gsm/GhoReserve.sol
```diff
@@ -142,9 +142,4 @@ contract GhoReserve is AccessControl, VersionedInitializable, IGhoReserve {
   function GHO_REMOTE_RESERVE_REVISION() public pure virtual override returns (uint256) {
     return 1;
```
avniculae: I think we can change this to a constant, to be aligned with FixedFeeStrategyFactory, wdyt?

src/contracts/facilitators/gsm/Gsm.sol
```diff
@@ -631,9 +632,4 @@ contract Gsm is AccessControl, VersionedInitializable, EIP712, IGsm {
 
     emit GhoReserveUpdated(oldReserve, newGhoReserve);
   }
-
-  /// @inheritdoc VersionedInitializable
-  function getRevision() internal pure virtual override returns (uint256) {
-    return GSM_REVISION();
```
avniculae, line 637: I'd convert GSM_REVISION to a constant to be aligned to FixedFeeStrategyFactory

tests/unit/TestGhoBase.t.sol
```diff
@@ -282,95 +282,59 @@ contract TestGhoBase is Test, Constants, Events {
...
-    address admin,
     address reserve
   ) internal returns (Gsm) {
-    Gsm gsmImpl = new Gsm(address(GHO_TOKEN), underlyingToken, priceStrategy);
-    AdminUpgradeabilityProxy gsmProxy = new AdminUpgradeabilityProxy(
-      address(gsmImpl),
+    Gsm gsm = new Gsm(address(GHO_TOKEN), underlyingToken, priceStrategy);
+    bytes memory initParams = abi.encodeWithSignature(
+      'initialize(address,address,uint128,address)',
+      address(this),
```
avniculae: I'd add a comment to mention this is the proxy admin

avniculae: although I think it would be better to use a random address, wdyt?

tests/unit/TestGhoBase.t.sol
avniculae, line 1: do we have tests to ensure the implementations of FixedFeeStrategyFactory, GhoReserve, Gsm, Gsm4626 are initialized after deployment?

src/contracts/facilitators/gsm/feeStrategy/FixedFeeStrategyFactory.sol
CheyenneAtapour: I think that makes sense if we want to continue this pattern of just deploying new ones each time. cc @miguelmtzinf wdyt?

tests/unit/TestGhoBase.t.sol
CheyenneAtapour: Added

tests/unit/TestGhoBase.t.sol
CheyenneAtapour: Done

src/contracts/facilitators/gsm/Gsm.sol
CheyenneAtapour: Done

src/contracts/facilitators/gsm/GhoReserve.sol
CheyenneAtapour: Done

src/contracts/facilitators/gsm/GhoReserve.sol
CheyenneAtapour: Good catch, done

src/contracts/facilitators/gsm/GhoReserve.sol
avniculae: I see you replaced `GHO_REMOTE_RESERVE_REVISION` with `REVISION`. I was thinking to keep the old name, to not introduce breaking changes, wdyt @miguelmtzinf ?

src/contracts/facilitators/gsm/Gsm.sol
avniculae: same as here https://github.com/aave-dao/gho-origin/pull/24#discussion_r2876919936

tests/unit/TestGhoGsmSteward.t.sol
```diff
@@ -44,6 +57,12 @@ contract TestGhoGsmSteward is TestGhoBase, GhoStewardProcedure {
     new GhoGsmSteward(address(0x001), address(0));
   }
 
+  function testRevertInitializeFixedFeeStrategyFactoryImplementation() public {
+    FixedFeeStrategyFactory factoryImpl = new FixedFeeStrategyFactory();
```
avniculae, line 76: additionally, we can also expectEmit event `emit Initialized(type(uint64).max);`

avniculae: applicable to the other tests of this kind as well

CheyenneAtapour: Done, but I put these in the testInitialize() function cause I think better suited there

avniculae:
This is not needed anymore since gho-origin updated aave-v3-origin to v3.6, which has a new version of VersionedInitializable (introduced in v3.4) that blocks initialize in constructor.

PR #34, opened 2026-03-04
GhoRouter

efecarranza:
# Changelog

Add GhoRouter contract to facilitate underlying -> GHO and back swaps using GSMs.
Allow for GHO -> sGHO and back as well in same contract.


src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,474 @@
...
+        if (allowed) {
+            _validateGsm(gsm);
+        }
+
+        isGsmAllowed[gsm] = allowed;
+        emit GsmAllowedUpdated(gsm, allowed);
+    }
+
+    /// @inheritdoc IGhoRouter
+    function previewSwapToGho(address gsm, address token, uint256 amount) external view returns (uint256, uint256) {
```
CheyenneAtapour: For this and swap functions, should we be validating the token? Here we only check the token is the stata token, but we still calculate the `previewDeposit` based on whatever amount was passed

CheyenneAtapour: The contract is, as documented, not supposed to hold user funds, so we can make that argument that swap is "safe", however, it's just a 1-liner to validate the actual token (vs just checking stata token or not), so I think perhaps worth it to add

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,474 @@
...
+
+    /// @inheritdoc IGhoRouter
+    function previewSwapFromGho(address gsm, address token, uint256 amount)
+        external
+        view
+        returns (uint256, uint256)
+    {
+        require(amount > 0, InvalidAmount());
+
+        return _previewTokenWithGho(gsm, token, amount);
```
CheyenneAtapour: As noted here: https://github.com/aave-dao/gho-origin/pull/34/changes#r2898432680, all these previews seem to have this issue of only checking if the token is the stata token, not really validating it. Thus these previews return inaccurate results for unrelated tokens passed as input

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,474 @@
...
+
+        require(IGsm(gsm).GHO_TOKEN() == GHO, InvalidGsm());
+        address stataToken = IGsm(gsm).UNDERLYING_ASSET();
+        require(stataToken != address(0), InvalidGsm());
+
+        require(IERC4626(stataToken).asset() != address(0), InvalidToken());
+    }
+
+    /**
+     * @dev Validates non zero inputs.
```
CheyenneAtapour:
```suggestion
     * @dev Validates non-zero inputs.
```
nit, the others are correctly like this already

tests/helpers/Constants.sol
```diff
@@ -57,6 +57,7 @@ contract Constants {
   uint128 constant DEFAULT_GSM_USDX_EXPOSURE = 100_000_000e6; // 6 decimals for USDX
   uint128 constant DEFAULT_GSM_USDX_AMOUNT = 100e6; // 6 decimals for USDX
   uint128 constant DEFAULT_GSM_GHO_AMOUNT = 100e18;
+  uint256 constant DEFAULT_SGHO_SUPPLY_CAP = 1_000_000 ether;
```
CheyenneAtapour: nit, inconsistency with the use of `ether` and `e18`

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,542 @@
...
+    function testConstructor() public {
+        GhoRouter router = new GhoRouter(address(this), address(GHO_TOKEN), address(SGHO));
+        assertEq(router.GHO(), address(GHO_TOKEN));
+        assertEq(router.sGHO(), address(SGHO));
+    }
+}
+
+contract GsmWhitelistTest is GhoRouterTest {
+    function testSetGsmAllowedNonOwner() public {
+        vm.startPrank(USER);
```
CheyenneAtapour: nit: could just be a `vm.prank`

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,542 @@
...
+        vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, USER));
+        GHO_ROUTER.setGsmAllowed(address(GHO_GSM_4626), false);
+        vm.stopPrank();
+    }
+
+    function testSetGsmAllowedZeroAddress() public {
+        vm.expectRevert(IGhoRouter.ZeroAddress.selector);
+        GHO_ROUTER.setGsmAllowed(address(0), false);
+        vm.stopPrank();
```
CheyenneAtapour: nit: what prank are we stopping here?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,490 @@
...
+        uint256 outputAmount = stataAmount;
+        if (token == stata) {
+            IERC20(stata).safeTransfer(outputReceiver, stataAmount);
+        } else {
+            outputAmount = IERC4626(stata).redeem(stataAmount, outputReceiver, address(this));
+        }
+        require(outputAmount >= minOutputAmount, SlippageExceeded());
+
+        if (ghoAmount > ghoSold) {
+            IERC20(GHO).safeTransfer(msg.sender, ghoAmount - ghoSold);
```
CheyenneAtapour: we refund the user, but it's possible we still have a dangling approval, since we approved `ghoAmount`, but it could be the case here that `ghoSold < ghoAmount`. Btw, I kinda like writing the condition this way I just wrote it, but nbd

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,490 @@
...
+            IERC20(token).forceApprove(stata, amount);
+            stataAmount = IERC4626(stata).deposit(amount, address(this));
+        }
+
+        IERC20(stata).forceApprove(gsm, stataAmount);
+        (uint256 assetSold, uint256 ghoAmount) = IGsm(gsm).sellAsset(stataAmount, address(this));
+
+        if (assetSold < stataAmount) {
+            if (token == stata) {
+                IERC20(stata).safeTransfer(msg.sender, stataAmount - assetSold);
```
CheyenneAtapour: In this case can also have dangling approval similar to https://github.com/aave-dao/gho-origin/pull/34/changes#r2921881033

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,490 @@
...
+
+    /// @inheritdoc IGhoRouter
+    function redeemSGho(uint256 sghoAmount, uint256 minOutputAmount, address recipient) external returns (uint256) {
+        return _redeemFromSGho(sghoAmount, minOutputAmount, recipient);
+    }
+
+    /// @inheritdoc IGhoRouter
+    function rescueToken(address token, address to, uint256 amount) external onlyOwner {
+        require(token != address(0), ZeroAddress());
+        require(to != address(0), ZeroAddress());
```
CheyenneAtapour: These 2 could be combined into one require. Additionally, do we want to ensure the amount is greater than 0?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,490 @@
...
+
+    /// @inheritdoc IGhoRouter
+    function swapFromGho(address gsm, address token, uint256 ghoAmount, uint256 minOutputAmount)
+        external
+        returns (uint256)
+    {
+        return _swapFromGho(gsm, token, ghoAmount, minOutputAmount, msg.sender);
+    }
+
+
```
CheyenneAtapour:
```suggestion
```
nit extra whitespace

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,490 @@
...
+    /// @inheritdoc IGhoRouter
+    function setGsmAllowed(address gsm, bool allowed) external onlyOwner {
+        require(gsm != address(0), ZeroAddress());
+
+        if (allowed) {
+            _validateGsm(gsm);
+        }
+
+        isGsmAllowed[gsm] = allowed;
```
CheyenneAtapour: nbd, but if we set this to the same value, we could revert instead of proceeding with this function

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        uint256 amount = 100 ether;
+
+        _dealAndApprove(address(GHO_TOKEN), address(GHO_ROUTER), amount);
+
+        // Set unreasonably high minOutputAmount to trigger slippage
+        vm.expectRevert(IGhoRouter.SlippageExceeded.selector);
+        vm.prank(USER);
+        GHO_ROUTER.swapFromGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), amount, type(uint256).max);
+
+        vm.stopPrank();
```
CheyenneAtapour:
```suggestion
```

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        GHO_ROUTER.swapFromGho(address(GHO_GSM_4626), address(WETH), amount, 0);
+        vm.stopPrank();
+    }
+
+    function testSwapFromGHOZeroAddressRecipient() public {
+        uint256 amount = 1 ether;
+        _dealAndApprove(address(GHO_TOKEN), address(GHO_ROUTER), amount);
+        vm.expectRevert(IGhoRouter.ZeroAddress.selector);
+        GHO_ROUTER.swapFromGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), amount, 0, address(0));
+        vm.stopPrank();
```
CheyenneAtapour:
```suggestion
```

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        SGHO.approve(address(GHO_ROUTER), amount);
+        vm.expectRevert(IGhoRouter.InvalidToken.selector);
+        GHO_ROUTER.swapFromSGho(address(GHO_GSM_4626), address(WETH), amount, 0);
+        vm.stopPrank();
+    }
+
+    function testSwapFromSGHOZeroAddressRecipient() public {
+        vm.expectRevert(IGhoRouter.ZeroAddress.selector);
+        GHO_ROUTER.swapFromSGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), 1 ether, 1 ether, address(0));
+        vm.stopPrank();
```
CheyenneAtapour:
```suggestion
```

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    function testSwapFromSGHOGSMNotAllowed() public {
+        GHO_ROUTER.setGsmAllowed(address(GHO_GSM_4626), false);
+        vm.expectRevert(IGhoRouter.GsmNotAllowed.selector);
+        GHO_ROUTER.swapFromSGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), 100 ether, 0);
+        vm.stopPrank();
```
CheyenneAtapour:
```suggestion
```

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    function testDepositForSGHOZeroAmount() public {
+        uint256 amount = 100 ether;
+
+        _dealAndApprove(address(GHO_TOKEN), address(GHO_ROUTER), amount);
+        vm.expectRevert(IGhoRouter.InvalidAmount.selector);
+        vm.startPrank(USER);
+        GHO_ROUTER.depositForSGho(0, amount + 1);
+        vm.stopPrank();
```
CheyenneAtapour: this can just be vm.prank()

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        uint256 amount = 100 ether;
+
+        _dealAndApprove(address(GHO_TOKEN), address(GHO_ROUTER), amount);
+        uint256 recipientBalanceBefore = IERC20(address(SGHO)).balanceOf(RECIPIENT);
+        uint256 userBalanceBefore = IERC20(address(SGHO)).balanceOf(USER);
+        vm.expectEmit(true, true, true, true, address(GHO_ROUTER));
+        emit IGhoRouter.SwapToSGho(USER, RECIPIENT, amount, amount);
+        vm.startPrank(USER);
+        uint256 shares = GHO_ROUTER.depositForSGho(amount, amount, RECIPIENT);
+        vm.stopPrank();
```
CheyenneAtapour: this can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        uint256 amount = 1000e6;
+        (uint256 ghoAmount, uint256 fee) = GHO_ROUTER.previewSwapToGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), amount);
+
+        assertGt(ghoAmount, 0, "Should preview GHO amount");
+        // Fee might be zero depending on GSM config, so we just check it doesn't revert
+        assertGe(fee, 0);
+    }
+
+    function testPreviewSwapToGHOInconsistentTokens() public {
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can just be vm.prank()

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+    function testPreviewSwapFromGHO() public view {
+        uint256 amount = 1_000 ether;
+        (uint256 outputAmount, uint256 fee) = GHO_ROUTER.previewSwapFromGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), amount);
+
+        assertGt(outputAmount, 0, "Should preview output amount");
+        assertGe(fee, 0);
+    }
+
+    function testPreviewSwapFromGHOInconsistentTokens() public {
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+    function testPreviewSwapToSGHO() public view {
+        uint256 amount = 1000 * 1e6;
+        (uint256 sghoAmount, uint256 fee) = GHO_ROUTER.previewSwapToSGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), amount);
+
+        assertGt(sghoAmount, 0, "Should preview sGHO amount");
+        assertGe(fee, 0, "Fee check should not revert");
+    }
+
+    function testPreviewSwapToSGHOInconsistentTokens() public {
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+    function testPreviewSwapFromSGHOUSDX_4626_TOKEN() public view {
+        uint256 amount = 100 ether;
+        (uint256 outputAmount, uint256 fee) = GHO_ROUTER.previewSwapFromSGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), amount);
+
+        assertGt(outputAmount, 0, "Should preview USDX_4626_TOKEN output");
+        assertGe(fee, 0, "Fee check should not revert");
+    }
+
+    function testPreviewSwapFromSGHOInconsistentTokens() public {
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        GHO_ROUTER.swapToGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), 0, 0);
+    }
+
+    function testSwapToGHOSlippageExceeded() public {
+        uint256 amount = 1000e6;
+        _dealAndApprove(address(USDX_4626_TOKEN), address(GHO_ROUTER), amount);
+
+        // Set unreasonably high minGHOAmount to trigger slippage
+        vm.expectRevert(IGhoRouter.SlippageExceeded.selector);
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can just be vm.prank()

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        GHO_ROUTER.swapToGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), amount, type(uint256).max);
+        vm.stopPrank();
+    }
+
+    function testSwapToGHOGSMNotAllowed() public {
+        GHO_ROUTER.setGsmAllowed(address(GHO_GSM_4626), false);
+
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can just be vm.prank()

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        deal(address(WETH), address(USER), 1);
+        WETH.approve(address(GHO_ROUTER), 1);
+        vm.expectRevert(IGhoRouter.InvalidToken.selector);
+        GHO_ROUTER.swapToGho(address(GHO_GSM_4626), address(WETH), 1, 0);
+        vm.stopPrank();
+    }
+
+    function testSwapToGHOZeroAddressRecipient() public {
+        vm.startPrank(USER);
```
CheyenneAtapour: this can be just vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        assertEq(ghoReceived, expectedOut, "Received GHO does not match");
+        assertEq(IERC20(GHO_TOKEN).balanceOf(RECIPIENT) - recipientBalanceBefore, ghoReceived, "Recipient gets GHO");
+        assertEq(IERC20(GHO_TOKEN).balanceOf(USER), userBalanceBefore, "Caller should not receive GHO");
+    }
+
+}
+
+contract SwapFromGHOTest is GhoRouterTest {
+    function testSwapFromGHOZeroAmount() public {
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        vm.prank(USER);
+        GHO_ROUTER.swapFromGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), amount, type(uint256).max);
+
+        vm.stopPrank();
+    }
+
+    function testSwapFromGHOGSMNotAllowed() public {
+        GHO_ROUTER.setGsmAllowed(address(GHO_GSM_4626), false);
+
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        vm.expectRevert(IGhoRouter.GsmNotAllowed.selector);
+        GHO_ROUTER.swapFromGho(address(GHO_GSM_4626), address(USDX_4626_TOKEN), 1, 0);
+        vm.stopPrank();
+    }
+
+    function testSwapFromGHOInconsistentTokens() public {
+        uint256 amount = 1 ether;
+        _dealAndApprove(address(GHO_TOKEN), address(GHO_ROUTER), amount);
+        vm.startPrank(USER);
```
CheyenneAtapour: this one can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        assertEq(IERC20(address(USDX_4626_TOKEN)).balanceOf(USER) - userBalanceBefore, outputAmount, "User gets static aToken");
+    }
+
+}
+
+contract SwapToSGhoTest is GhoRouterTest {
+    function testSwapToSGHOGSMNotAllowed() public {
+        GHO_ROUTER.setGsmAllowed(address(GHO_GSM_4626), false);
+
+        vm.startPrank(USER);
```
CheyenneAtapour: this can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+
+        assertEq(shares, expectedOut, "Should receive sGHO shares");
+        assertEq(IERC20(address(SGHO)).balanceOf(USER), 0, "User should receive zero shares");
+        assertEq(IERC20(address(SGHO)).balanceOf(RECIPIENT), shares, "Recipient should receive minted shares");
+    }
+}
+
+contract SwapFromSGhoTest is GhoRouterTest {
+    function testSwapFromSGHOZeroAmount() public {
+        vm.startPrank(USER);
```
CheyenneAtapour: this can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+}
+
+contract DepositForSGhoTest is GhoRouterTest {
+    function testDepositForSGHO() public {
+        uint256 amount = 100 ether;
+
+        _dealAndApprove(address(GHO_TOKEN), address(GHO_ROUTER), amount);
+        vm.expectEmit(true, true, true, true, address(GHO_ROUTER));
+        emit IGhoRouter.SwapToSGho(USER, USER, amount, amount);
+        vm.startPrank(USER);
```
CheyenneAtapour: this can just be vm.prank()

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        assertEq(shares, amount, "sGHO copy should mint 1:1 shares at initial index");
+        assertEq(IERC20(address(SGHO)).balanceOf(USER), amount, "User should receive all shares");
+    }
+
+    function testDepositForSGHOSlippageExceeded() public {
+        uint256 amount = 100 ether;
+
+        _dealAndApprove(address(GHO_TOKEN), address(GHO_ROUTER), amount);
+        vm.expectRevert(IGhoRouter.SlippageExceeded.selector);
+        vm.startPrank(USER);
```
CheyenneAtapour: This can just be vm.prank

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+        GHO_ROUTER.depositForSGho(amount, amount + 1);
+        vm.stopPrank();
+    }
+
+    function testDepositForSGHOZeroAmount() public {
+        uint256 amount = 100 ether;
+
+        _dealAndApprove(address(GHO_TOKEN), address(GHO_ROUTER), amount);
+        vm.expectRevert(IGhoRouter.InvalidAmount.selector);
+        vm.startPrank(USER);
```
CheyenneAtapour: this can just be vm.prank()

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,682 @@
...
+    function testSwapToSGHOInconsistentTokens() public {
+        vm.startPrank(USER);
+        deal(address(WETH), address(USER), 1);
+        WETH.approve(address(GHO_ROUTER), 1);
+        vm.expectRevert(IGhoRouter.InvalidToken.selector);
+        GHO_ROUTER.swapToSGho(address(GHO_GSM_4626), address(WETH), 1, 0);
+        vm.stopPrank();
+    }
+
+    function testSwapToSGHOZeroAddressRecipient() public {
```
CheyenneAtapour: this name is wrong, it actually tests: `depositForSGho`

tests/unit/TestGhoRouter.t.sol
CheyenneAtapour, line 1: What about testing the following: 
* swapToGho with stata token, normal case
* swapToSGho with stata token
* swapFromSGho with a valid recipient address passed as param
* rescueToken called by non-owner
* Failure cases for _validateGsm() (via testing setGsmAllowed() with an invalid gsm contract, can mock)
* Failure cases for preview when previewing amount 0
* Preview tests with underlying token instead of stata

Also currently no fuzz tests, but perhaps that's fine because a lot of the logic that we would fuzz in a test is covered by whatever the actual gsm would be doing

docs/gho-router.md
```diff
@@ -0,0 +1,267 @@
+# GhoRouter Failure Modes and Security Assumptions
+
+This document captures trust assumptions and failure modes for the current
+router implementation at:
+
+- `src/GhoRouter.sol`
```
CheyenneAtapour:
```suggestion
- `src/contracts/misc/GhoRouter.sol`
```

docs/gho-router.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Scope and trust boundary
+
+`GhoRouter` orchestrates six flows:
+
+- `swapToGho(gsm, token, amount, minGHOAmount[, recipient])` for GSM token -> GHO
+- `swapFromGho(gsm, token, ghoAmount, minOutputAmount[, recipient])` for GHO -> GSM underlying token/static aToken
+- `swapToSGho(gsm, token, amount, minSGHOAmount[, recipient])` for GSM token -> sGHO
+- `depositForSGho(ghoAmount, minSGHOAmount[, recipient])` for direct GHO -> sGHO
+- `swapFromsGHO(gsm, token, sghoAmount, minOutputAmount[, recipient])` for sGHO -> GSM underlying token/static aToken
```
CheyenneAtapour:
```suggestion
- `swapFromSGHO(gsm, token, sghoAmount, minOutputAmount[, recipient])` for sGHO -> GSM underlying token/static aToken
```

src/contracts/misc/GhoRouter.sol
CheyenneAtapour: Ah, I see you have commented the dangling approval issue in readme, but perhaps worth considering removing

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,490 @@
...
+        emit GsmAllowedUpdated(gsm, allowed);
+    }
+
+    /// @inheritdoc IGhoRouter
+    function previewSwapToGho(address gsm, address token, uint256 amount) external view returns (uint256, uint256) {
+        require(amount > 0, InvalidAmount());
+
+        address stata = IGsm(gsm).UNDERLYING_ASSET();
+        require(token == stata || token == IERC4626(stata).asset(), InvalidToken());
```
CheyenneAtapour: Maybe could be useful to have a helper to validate the token instead of copy pasting? But not really a huge deal

src/contracts/misc/GhoRouter.sol
efecarranza: don't think it's really necessary and was trying to save gas but i've added it for completeness. even mainnet is cheap nowdays.

src/contracts/misc/GhoRouter.sol
efecarranza: same as above

src/contracts/misc/GhoRouter.sol
efecarranza: it's an onlyOwner function, to me just the `to` check would be enough since token being 0x0 would revert, wouldn't add the amount > 0 check

src/contracts/misc/GhoRouter.sol
efecarranza: i don't think it's necessary. myself

src/contracts/misc/GhoRouter.sol
CheyenneAtapour: You've added one, but not in the else clause when token is not the stata token. If you want to go for completeness, we need to add it in this case too

tests/unit/TestGhoRouter.t.sol
CheyenneAtapour, line 1: Can add a swapFromGho with recipient test, seems like that case is missing

docs/gho-router.md
```diff
@@ -0,0 +1,267 @@
...
+- `_validateGsm` checks compatibility when allowlisting is enabled, not on every swap.
+- Preview methods also do not enforce `isGsmAllowed`.
+
+## Security assumptions
+
+The current implementation assumes:
+
+1. Owner securely manages `isGsmAllowed` and curates safe GSM addresses.
+2. `IGSM`, `IStaticAToken`, and `sGHO` dependencies are interface-compatible and non-malicious.
+3. `sGHO` is a valid ERC4626 vault over GHO (constructor does not enforce `asset() == GHO`).
```
CheyenneAtapour: It would be easy to just add this check, right?

docs/gho-router.md
```diff
@@ -0,0 +1,267 @@
...
+- `rescueToken` is `onlyOwner`.
+
+Limitations of current implementation:
+
+- No internal pause/circuit-breaker.
+- No explicit reentrancy guard.
+- No on-swap revalidation of GSM `GHO_TOKEN`.
+- Preview methods can return quotes for non-allowlisted GSMs.
+- The router does not explicitly reset allowances to zero after each operation.
+- On GSM buy paths, if `buyAsset` spends less than transferred GHO, leftover GHO can remain in the router.
```
CheyenneAtapour: This is now solved for with the refunding, right?

docs/gho-router.md
```diff
@@ -0,0 +1,267 @@
...
+
+- Quote UX mismatch, revert, or lower output.
+
+Current mitigation:
+
+- Execution-time minimum output checks.
+- Conservative slippage buffers.
+- Integrators should check `isGsmAllowed(gsm)` before submitting GSM swaps.
+
+### 5) Partial consumption and residual balances
```
CheyenneAtapour, line 195: I think this has been resolved

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,494 @@
...
+     * @return ghoAmount Amount of GHO received from GSM.
+     */
+    function _sellTokenForGho(address gsm, address token, uint256 amount, uint256 minGhoAmount)
+        internal
+        returns (uint256, uint256)
+    {
+        address stata = IGsm(gsm).UNDERLYING_ASSET();
+        uint256 stataAmount = amount;
+        if (token != stata) {
+            require(token == IERC4626(stata).asset(), InvalidToken());
```
CheyenneAtapour: Could use the helper here to validate the token, so this is done consistently across functions

docs/gho-router.md
CheyenneAtapour: Now that we have this, can remove this assumption

docs/gho-router.md
```diff
@@ -0,0 +1,267 @@
...
+  - If `sellAsset` consumes less than requested shares, remaining shares are redeemed to `msg.sender`.
+- `rescueToken` is `onlyOwner`.
+
+Limitations of current implementation:
+
+- No internal pause/circuit-breaker.
+- No explicit reentrancy guard.
+- No on-swap revalidation of GSM `GHO_TOKEN`.
+- Preview methods can return quotes for non-allowlisted GSMs.
+- The router does not explicitly reset allowances to zero after each operation.
```
CheyenneAtapour, line 117: If we complete the dangling approval fix, can update this

docs/gho-router.md
```diff
@@ -0,0 +1,267 @@
...
+
+### 8) Residual allowance risk
+
+What can happen:
+
+- If funds are stranded in the router, approved spenders may retain pull capability.
+
+Why:
+
+- Allowances set via `forceApprove` are not explicitly cleared after calls.
```
CheyenneAtapour, line 259: If we fix, can remove

docs/gho-router.md
```diff
@@ -0,0 +1,267 @@
...
+
+Current mitigation:
+
+- Slippage checks on outputs.
+- Owner-operated `rescueToken` for recovery.
+- Balance monitoring by operators.
+- Refund mechanism.
+- Set approvals to zero.
+
+### 6) sGHO compatibility or misconfiguration
```
CheyenneAtapour, line 215: This section is now outdated since the constructor does check the underlying is GHO

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,495 @@
...
+        (uint256 assetSold, uint256 ghoAmount) = IGsm(gsm).sellAsset(stataAmount, address(this));
+
+        if (assetSold < stataAmount) {
+            if (token == stata) {
+                IERC20(stata).safeTransfer(msg.sender, stataAmount - assetSold);
+                IERC20(stata).forceApprove(gsm, 0);
+            } else {
+                uint256 redeemed = IERC4626(stata).redeem(stataAmount - assetSold, msg.sender, address(this));
+                amount = amount - redeemed;
+            }
```
CheyenneAtapour:
```suggestion
            } else {
                uint256 redeemed = IERC4626(stata).redeem(stataAmount - assetSold, msg.sender, address(this));
                amount = amount - redeemed;
            }
            IERC20(stata).forceApprove(gsm, 0);
```

src/contracts/misc/interfaces/IGhoRouter.sol
```diff
@@ -28,30 +28,57 @@ interface IGhoRouter {
...
      * @param outputAmount The amount of output tokens received
      */
-    event SwapFromGho(address indexed user, address indexed recipient, uint256 ghoAmount, uint256 outputAmount);
+    event SwapFromGho(address indexed user, address indexed recipient, address indexed token, uint256 ghoAmount, uint256 outputAmount);
 
     /**
      * @notice Emitted when a swap to sGHO is completed
      * @param user The address of the user who initiated the swap
      * @param recipient The address of the recipient of the acquired token
+     * * @param token The address of the token that was swapped
```
CheyenneAtapour: extra asterisk here

src/contracts/misc/interfaces/IGhoRouter.sol
```diff
@@ -60,12 +87,14 @@ interface IGhoRouter {
      * @notice Emitted when a swap from sGHO is completed
      * @param user The address of the user who initiated the swap
      * @param recipient The address of the recipient of the acquired token
+     * @param token The address of the token that was swapped to
```
CheyenneAtapour: not a big deal, but you could say address of "output token". Also more consistent with how you refer to it below

docs/gho-router.md
CheyenneAtapour: remove this

CheyenneAtapour: It currently does more than this, does check underlying is gho

CheyenneAtapour (approved):

Kogaroshi (commented):
While I think the internal logic if the Router is good as is, I think there is too many entry point for what a Router should actually be. The goal here should be to abstract away all the complexity of the route swaps.
We could have a struct for each asset/GSM pair (which would also need to have a flag to know if it's a simple GSM or one supporting stataTokens), so we can only have 2 entry points (or 4 if we consider the variants with `receiver`) : 
swapToGho & swapFromGho, that only take as parameters the `asset`, the exact amount in, and the min output amount.
This would reduce the complexity for integrators, as they would have simple preview methods and swap methods.

tests/unit/TestGhoRouter.t.sol
```diff
@@ -0,0 +1,859 @@
...
+    return address(0);
+  }
+}
+
+/**
+ * @title GhoRouterTest
+ * @notice Integration tests for GhoRouter on mainnet fork
+ * @dev Run with: forge test --match-path test/fork/onboarding/GhoRouterTest.t.sol -vvv
+ */
+contract GhoRouterTest is TestGhoBase {
```
Kogaroshi: Would be cleaner to have all those test suites as different files of the `tests/unit/GhoRouter` dir, instead of a single file

tests/unit/TestGhoRouter.t.sol
Kogaroshi: @efecarranza pls add fuzz tests for all swap and preview actions, confirming consistency with sGHO & GSM fetched values for all routes.

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+import {IGsm} from 'src/contracts/facilitators/gsm/interfaces/IGsm.sol';
+import {IGhoRouter} from 'src/contracts/misc/interfaces/IGhoRouter.sol';
+
+/**
+ * @title GhoRouter
+ * @author TokenLogic
+ * @notice Router for token swaps through whitelisted GSMs and direct GHO/sGHO conversion paths
+ * @dev This contract never stores user funds and uses exact approvals only
+ */
+contract GhoRouter is Ownable, IGhoRouter {
```
avniculae: Should we use Ownable2Step?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+  /// @inheritdoc IGhoRouter
+  mapping(address token => address gsm) public gsms;
+
+  /**
+   * @dev Constructor to initialize the contract
+   * @param initialOwner Address of the contract owner
+   * @param gho Address of the GHO token on the deployed network
+   * @param sgho Address of sGHO on the deployed network
+   */
+  constructor(address initialOwner, address gho, address sgho) Ownable(initialOwner) {
```
avniculae, line 41: nit: we should use `_` suffix for all vars in constructor: `initialOwner_`, `gho_`, `sGho_`

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+      } else {
+        amountOut = _swapFromSGho(tokenOut, exactAmountIn, minAmountOut, recipient);
+      }
+    } else {
+      if (tokenOut == GHO) {
+        amountOut = _swapToGho(tokenIn, exactAmountIn, minAmountOut, recipient);
+      } else if (tokenOut == sGHO) {
+        amountOut = _swapToSGho(tokenIn, exactAmountIn, minAmountOut, recipient);
+      } else {
+        revert InvalidPath();
```
avniculae: I'd rename this to `InvalidTokenPair` or smth similar, and I would emit tokenIn and tokenOut, wdyt?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+      }
+    }
+
+    emit Swap(tokenIn, tokenOut, exactAmountIn, amountOut, recipient);
+
+    return amountOut;
+  }
+
+  /// @inheritdoc IGhoRouter
+  function setTokenToGsm(address token, address gsm) external onlyOwner {
```
avniculae: I'd rename this to `setTokenGsm`, wdyt? error name can also be changed to `TokenGsmAlreadySet`, and event to `TokenGsmAdded`

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+    require(gsms[token] == address(0), TokenToGsmAlreadySet());
+
+    _validateGsm(gsm, token);
+
+    gsms[token] = gsm;
+    emit TokenToGsmAdded(token, gsm);
+  }
+
+  /// @inheritdoc IGhoRouter
+  function removeTokenToGsm(address token) external onlyOwner {
```
avniculae: similar, I'd remove `To` from namings, wdyt?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+   */
+  function _previewSwapToSGho(address token, uint256 amount) internal view returns (uint256) {
+    address gsm = gsms[token];
+    require(gsm != address(0), GsmNotConfigured());
+
+    address stata = IGsm(gsm).UNDERLYING_ASSET();
+
+    _validateTokens(token, stata);
+    uint256 sharesAmount = token == stata ? amount : IERC4626(stata).previewDeposit(amount);
+    (, uint256 ghoAmount, , ) = IGsm(gsm).getGhoAmountForSellAsset(sharesAmount);
```
avniculae: all of this can be replaced with `_previewSwapToGho()`, no?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+    uint256 exactAmountIn,
+    uint256 minSGHOAmount,
+    address recipient
+  ) internal returns (uint256) {
+    address gsm = gsms[token];
+    require(gsm != address(0), GsmNotConfigured());
+
+    IERC20(token).safeTransferFrom(msg.sender, address(this), exactAmountIn);
+
+    (, uint256 ghoAmount) = _sellTokenForGho(gsm, token, exactAmountIn, 0);
```
avniculae: this part is exactly _swapToGho apart from the transfer. wondering if we can reuse code in a clean way, wdyt

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+  function _swapFromSGho(
+    address token,
+    uint256 exactAmountIn,
+    uint256 minOutputAmount,
+    address recipient
+  ) internal returns (uint256) {
+    address gsm = gsms[token];
+    require(gsm != address(0), GsmNotConfigured());
+
+    uint256 ghoAmount = _redeemGho(exactAmountIn, 0);
```
avniculae, line 343: apart form this redeem this function is the same as _swapFromGho(), wondering if we could reuse code

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+    require(exactAmountIn > 0, InvalidAmount());
+
+    uint256 amountOut;
+    if (tokenIn == GHO) {
+      if (tokenOut == sGHO) {
+        amountOut = IERC4626(sGHO).previewDeposit(exactAmountIn);
+      } else {
+        amountOut = _previewTokenWithGho(tokenOut, exactAmountIn);
+      }
+    } else if (tokenIn == sGHO) {
```
avniculae, line 168: nit: in this if statement you always convert from sgho to gho, so we could do

```solidity
amountOut = _previewSwapToGho(..)
if (tokenOut != GHO) {
  amountOut = _previewTokenWithGho(..)
}
```

avniculae: also to me previewTokenWithGho is not intuitive. I'd have prefered to do `preview{WriteAction}`, where WriteAction is the name of the fn that is called in "swap"

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+    uint256 deadline
+  ) external returns (uint256) {
+    _validateInputs(exactAmountIn, recipient, deadline);
+
+    uint256 amountOut;
+    if (tokenIn == GHO) {
+      if (tokenOut == sGHO) {
+        amountOut = _depositForSGho(exactAmountIn, minAmountOut, recipient);
+      } else {
+        amountOut = _swapFromGho(tokenOut, exactAmountIn, minAmountOut, recipient);
```
avniculae: fn naming is a bit hard to follow for me, I'm wondering if we could use different names that are easier to understand:

eg underlyingToGho, GhoToUnderlying, ghoToSgho, sGhoToGho, sGhoToUnderlying etc.

tests/unit/GhoRouter/TestGhoRouterSetTokenToGsm.t.sol
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  function testSetTokenToGsm() public {
+    // Remove previously set GSM on setUp
+    GHO_ROUTER.removeTokenToGsm(address(USDX_TOKEN));
+    assertEq(GHO_ROUTER.gsms(address(USDX_TOKEN)), address(0));
+
+    vm.expectEmit(true, true, true, true, address(GHO_ROUTER));
+    emit IGhoRouter.TokenToGsmAdded(address(USDX_TOKEN), address(GHO_GSM_4626));
+    GHO_ROUTER.setTokenToGsm(address(USDX_TOKEN), address(GHO_GSM_4626));
```
avniculae: I don't think we have tests where we set gsm to non-4626 gsm, correct? If yes, let's add

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+      if (tokenOut == GHO) {
+        amountOut = _swapToGho(tokenIn, exactAmountIn, minAmountOut, recipient);
+      } else if (tokenOut == sGHO) {
+        amountOut = _swapToSGho(tokenIn, exactAmountIn, minAmountOut, recipient);
+      } else {
+        revert InvalidPath();
+      }
+    }
+
+    emit Swap(tokenIn, tokenOut, exactAmountIn, amountOut, recipient);
```
avniculae: _swapFromGho may transfer less than `exactAmountIn`, is it indented to emit `exactAmountIn` here?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+    address stata = IGsm(gsm).UNDERLYING_ASSET();
+    uint256 stataAmount = exactAmountIn;
+    if (token != stata) {
+      _validateTokens(token, stata);
+      IERC20(token).forceApprove(stata, exactAmountIn);
+      stataAmount = IERC4626(stata).deposit(exactAmountIn, address(this));
+    }
+
+    IERC20(stata).forceApprove(gsm, stataAmount);
+    (uint256 assetSold, uint256 ghoAmount) = IGsm(gsm).sellAsset(stataAmount, address(this));
```
avniculae: shouldn't we forceApprove 0 after this? may be applicable in other places as well

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,491 @@
...
+
+    uint256 ghoAmount = _redeemGho(exactAmountIn, 0);
+
+    (uint256 stataAmountToBuy, uint256 ghoUsed, , ) = IGsm(gsm).getAssetAmountForBuyAsset(
+      ghoAmount
+    );
+    (, uint256 outputAmount) = _buyTokenWithGho(
+      gsm,
+      token,
+      ghoUsed,
```
avniculae: if ghoAmount is different than ghoUsed, we'll leave dust gho in this router. I think we should fix this, wdyt?

src/contracts/misc/GhoRouter.sol
efecarranza: i followed the convention of the rest of the repo

src/contracts/misc/GhoRouter.sol
efecarranza: i followed the convention of the rest of the repo

src/contracts/misc/GhoRouter.sol
efecarranza: done

src/contracts/misc/GhoRouter.sol
efecarranza: with the audit having been performed and on the changes stage, i'll keep name changes. apologies!

src/contracts/misc/GhoRouter.sol
efecarranza: with the audit having been performed and on the changes stage, i'll keep name changes. apologies!

src/contracts/misc/GhoRouter.sol
efecarranza: yep, thanks!

src/contracts/misc/GhoRouter.sol
efecarranza: done

src/contracts/misc/GhoRouter.sol
efecarranza: hmm, for this one i couldn't find a good case because of the transfers in that are different + needing the gho used

src/contracts/misc/GhoRouter.sol
avniculae: We are pushing Ownable2Step for all new contracts, so I think it would be good to add it if not inconvenient audit-wise

src/contracts/misc/GhoRouter.sol
avniculae: I think we can use the new conventions for the new contracts as a general rule

src/contracts/misc/GhoRouter.sol
avniculae: wdym regarding ghoUsed? that's the same here and in swapFromGho right? We just need to handle the transfer/redeemgho difference correct?

src/contracts/misc/GhoRouter.sol
efecarranza: it's a minimal amount that could be left over (in wei). please see the report

src/contracts/misc/GhoRouter.sol
efecarranza: updated so it's always exactAmountIn for the user

src/contracts/misc/GhoRouter.sol
efecarranza: posted report below, it's more expensive in gas terms to do something about the dust than to let it accumulate. prior version had refund mechanisms/dust and audit suggested we remove

tests/unit/GhoRouter/TestGhoRouterSetTokenToGsm.t.sol
efecarranza: added test to show this works with regular GSMs just as well

src/contracts/misc/GhoRouter.sol
avniculae: what do you mean?

tests/unit/GhoRouter/TestGhoRouterSetTokenToGsm.t.sol
avniculae: I don't see the new tests, can you pls point me to the commit? We need tests for non 4626 for all flows right?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+   * @param exactAmountIn Amount of input tokens pulled from the caller.
+   * @param minGhoAmount Minimum acceptable GHO output.
+   * @return ghoAmount Amount of GHO received from GSM.
+   */
+  function _performSwapToGho(
+    address token,
+    uint256 exactAmountIn,
+    uint256 minGhoAmount
+  ) internal returns (uint256) {
+    address gsm = gsms[token];
```
avniculae: - if we want to use gsm4626 of USDC, then we need to register the gsm for both stata usdc token, and for  usdc token (since later we detect and wrap usdc into stata token).
- however, bc of this mechanism, if we register non 4626 for USDC, then we cannot use gho router to wrap usdc to stata usdc anymore

thoughts?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+    return (token == stata ? assetSold : exactAmountIn, ghoAmount);
+  }
+
+  /**
+   * @dev Buys GSM static aTokens with GHO, then returns either static or underlying output based on `token`.
+   * @dev An almost negligible amount of dust can be left unsued, would cost more in gas than amount returned.
+   * @param gsm Whitelisted GSM used for the buy path.
+   * @param token Output token requested by the caller (underlying token or static aToken).
+   * @param exactAmountIn GHO budget used to acquire static aTokens.
+   * @param stataAmountToBuy Amount of stata token to acquire from the GSM.
```
avniculae: this is not necessarily correct right? if gsm is non 4626 one, then this is false I think

avniculae: I general we use stata terminology even if token is not static a token, which I think it can be confusing to readers and devs

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+    address gsm = gsms[token];
+    require(gsm != address(0), GsmNotConfigured());
+
+    address stata = IGsm(gsm).UNDERLYING_ASSET();
+
+    _validateTokens(token, stata);
+    (uint256 assetAmount, , , ) = IGsm(gsm).getAssetAmountForBuyAsset(ghoAmount);
+    uint256 outputAmount = token == stata
+      ? assetAmount
+      : IERC4626(stata).previewRedeem(assetAmount);
```
avniculae: asset amount here is shares of the static a token, which is fine, just wondering whether redeem(assetAmount) is confusing as redeem takes shares, not assets (from erc4626 pov)

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+  }
+
+  /**
+   * @dev Pulls sGHO from the caller and redeems it for GHO into the router.
+   * @param exactAmountIn Amount of sGHO shares to redeem.
+   * @param minGhoAmount Minimum acceptable GHO output.
+   * @return ghoAmount Amount of GHO redeemed.
+   */
+  function _redeemGho(uint256 exactAmountIn, uint256 minGhoAmount) internal returns (uint256) {
+    uint256 ghoAmount = IERC4626(sGHO).redeem(exactAmountIn, address(this), msg.sender);
```
avniculae: I'd use named args here an in all similar fn calls

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+  }
+
+  /**
+   * @dev Redeems sGHO directly into GHO and forwards output to the recipient.
+   * @param exactAmountIn Amount of sGHO shares pulled from the caller.
+   * @param minOutputAmount Minimum acceptable GHO output.
+   * @param recipient Address that receives the resulting GHO.
+   * @return ghoAmount Amount of GHO sent to `recipient`.
+   */
+  function _redeemFromSGho(
```
avniculae: it's very hard for readers to understand difference between `_redeemGho` and `_redeemFromSGho`. I know audits have been done, but I'd be in favour of renaming these to better names, as previously suggested: ghoToSGho, sGhoToGho, sGhoToGhoWithTransfer, etc

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+  ) internal returns (uint256) {
+    IERC20(GHO).safeTransferFrom(msg.sender, address(this), exactAmountIn);
+    uint256 sghoAmount = _depositGho(exactAmountIn, recipient, minSGHOAmount);
+
+    return sghoAmount;
+  }
+
+  /**
+   * @dev Redeems sGHO into GHO, then swaps through GSM into a caller-selected token.
+   * @dev An almost negligible amount of dust can be left unsued, would cost more in gas than amount returned.
```
avniculae, line 326: I don't think gas should be a reason here, I'd like us to sweep everything that may be left behind, it would be a strong invariant

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+  ) internal returns (uint256) {
+    address gsm = gsms[token];
+    require(gsm != address(0), GsmNotConfigured());
+
+    uint256 ghoAmount = _redeemGho(exactAmountIn, 0);
+
+    (uint256 stataAmountToBuy, uint256 ghoUsed, , ) = IGsm(gsm).getAssetAmountForBuyAsset(
+      ghoAmount
+    );
+    (, uint256 outputAmount) = _buyTokenWithGho(
```
avniculae: let's use named args here and in every similar fn

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+    address recipient
+  ) internal returns (uint256) {
+    address gsm = gsms[token];
+    require(gsm != address(0), GsmNotConfigured());
+
+    (uint256 stataAmountToBuy, uint256 ghoUsed, , ) = IGsm(gsm).getAssetAmountForBuyAsset(
+      exactAmountIn
+    );
+
+    IERC20(GHO).safeTransferFrom(msg.sender, address(this), exactAmountIn);
```
avniculae: why do we transfer exactAmountIn instead of ghoUsed? to make sure no dust is left in user's wallet?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+    _validateTokens(token, stata);
+    uint256 sharesAmount = token == stata ? amount : IERC4626(stata).previewDeposit(amount);
+    (, uint256 ghoAmount, , ) = IGsm(gsm).getGhoAmountForSellAsset(sharesAmount);
+
+    return ghoAmount;
+  }
+
+  /**
+   * @notice Preview the amount of sGHO received for a given input amount through a GSM path
+   * @dev This is an estimation and actual results may vary slightly due to interest accrual
```
avniculae: I'd remove this. if called in the same tx, then it's not an estimation

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,501 @@
...
+        revert InvalidTokenPair(tokenIn, tokenOut);
+      }
+    }
+
+    return amountOut;
+  }
+
+  /**
+   * @notice Preview the amount of GHO received for a given input amount
+   * @dev This is an estimation and actual results may vary slightly due to interest accrual
```
avniculae: same as https://github.com/aave-dao/gho-origin/pull/34/changes#r3373452190

src/contracts/misc/GhoRouter.sol
efecarranza: in sell token there's no dust

src/contracts/misc/GhoRouter.sol
Kogaroshi: We might even want to use `Ownable2StepWithGuardian` to have a Pause guardian here, allowing to pause this Router if we notice any issue during swaps

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,567 @@
...
+  function _ghoToToken(
+    address token,
+    address gsm,
+    uint256 exactAmountIn,
+    uint256 minOutputAmount,
+    address recipient
+  ) internal returns (uint256, uint256) {
+    (address gsmAsset, bool isStata) = _validateGsmInput(token, gsm);
+
+    (uint256 amountToBuy, uint256 ghoUsed, , ) = IGsm(gsm).getAssetAmountForBuyAsset(exactAmountIn);
```
Kogaroshi, line 266: we never verify that `ghoUsed <= exactAmountIn`. We might trust current GSMs, but we cannot be sure about future ones.

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,567 @@
...
+
+    return ghoAmount;
+  }
+
+  /**
+   * @dev Swaps GHO into a caller-selected GSM token (underlying or static aToken) and forwards output.
+   * @dev An almost negligible amount of dust can be left unsued, would cost more in gas than amount returned.
+   * @param token Output token (either GSM underlying token or its static aToken).
+   * @param gsm Whitelisted GSM used for the swap.
+   * @param exactAmountIn Amount of GHO pulled from the caller.
```
Kogaroshi: Not true, we pull `ghoUsed` here

src/contracts/misc/GhoRouter.sol
avniculae: Addressed: the token/GSM pairing validation is now centralized in `_validateGsmInput`, used by all swap and preview paths.

tests/unit/TestGhoRouter.t.sol
avniculae: Added: `testSwapFromGHOToStataUSDX_4626_TOKENWithRecipient` covers the GHO -> token flow with a separate recipient.

docs/gho-router.md
avniculae: The constructor now enforces `IERC4626(sgho).asset() == gho`, and the docs assumption was updated to reflect that.

docs/gho-router.md
avniculae: The refund mechanism was later removed based on audit feedback; instead the buy path now pulls only the GHO actually used (`ghoUsed`) from the caller, so there is nothing to refund on GHO -> token. Residual GHO is only possible on the sGHO -> token path and is documented under failure mode 5.

src/contracts/misc/GhoRouter.sol
avniculae: Resolved by extracting the shared buy logic into `_buyTokenWithGho`; `_ghoToToken` and `_sGhoToToken` now only differ in how the GHO is sourced (transferFrom vs sGHO redeem).

src/contracts/misc/GhoRouter.sol
avniculae: The preview helpers were renamed to mirror the write actions: `_previewGhoToToken`, `_previewTokenToGho`, `_previewTokenToSGho`, matching `_ghoToToken` / `_tokenToGho` / `_tokenToSGho`.

src/contracts/misc/GhoRouter.sol
avniculae: Done — the internal flows are now named `_ghoToSGho`, `_ghoToToken`, `_sGhoToGho`, `_sGhoToToken`, `_tokenToGho`, `_tokenToSGho`.

src/contracts/misc/GhoRouter.sol
avniculae: Resolved: `_ghoToToken` now returns `ghoUsed`, `swap()` reassigns `exactAmountIn` to it before emitting, and the router only pulls `ghoUsed` from the caller — so the event reflects the GHO actually spent (covered by `testSwapFromGHOToUSDX_4626_TOKEN_GSMUsesLessGho`).

src/contracts/misc/GhoRouter.sol
avniculae: Resolved by the redesign in 7cb687d: the GSM allowlist is decoupled from tokens — the caller supplies the `gsm` per swap and wrapping is driven by the curated `tokenToStata` registry, so a plain GSM and a stata GSM for the same underlying can coexist.

src/contracts/misc/GhoRouter.sol
avniculae: Resolved: whether wrapping is needed is now derived from the `tokenToStata` registry in `_validateGsmInput` (`isStata` is false for plain GSM assets), and the misleading `stata` naming was replaced with `gsmAsset`.

src/contracts/misc/GhoRouter.sol
avniculae: Now uses named args (`redeem({shares: ...})`), making the shares semantics explicit.

src/contracts/misc/GhoRouter.sol
avniculae: Done — named args are used here and on the other deposit/redeem/sellAsset/buyAsset calls.

src/contracts/misc/GhoRouter.sol
avniculae: Done — named args are used here and in the similar calls.

src/contracts/misc/GhoRouter.sol
avniculae: Renamed as suggested: the flows are now `_ghoToSGho`, `_sGhoToGho`, `_sGhoToToken`, etc., with only the low-level `_depositGho` / `_redeemGho` helpers kept.

src/contracts/misc/GhoRouter.sol
avniculae: Changed: the router now pulls only `ghoUsed` from the caller.

tests/unit/GhoRouter/TestGhoRouterSwap.t.sol
```diff
@@ -0,0 +1,898 @@
...
+    assertEq(
+      IERC20(address(SGHO)).balanceOf(RECIPIENT),
+      actualShares,
+      'Recipient should receive minted shares'
+    );
+  }
+}
+
+contract SwapFromSGhoTest is TestGhoRouterBase {
+  function testSwapFromSGHOZeroAmount() public {
```
avniculae, line 595: Copy-paste: this swaps GHO -> USDX_4626_TOKEN, not sGHO. It passes because `_validateInputs` runs before routing, but the sGHO zero-amount case isn't actually covered.

tests/unit/GhoRouter/TestGhoRouterPreviewSwap.t.sol
```diff
@@ -0,0 +1,187 @@
...
+    assertEq(outputAmount, expectedOut, 'Should preview output amount');
+  }
+
+  function testPreviewSwapFromGHOInconsistentTokens() public {
+    vm.prank(USER);
+    vm.expectRevert(IGhoRouter.InvalidToken.selector);
+    GHO_ROUTER.previewSwap(address(GHO_TOKEN), address(WETH), address(GHO_GSM_4626), 1);
+  }
+
+  function testPreviewSwapFromGHOToStataUSDX_4626_TOKEN(uint256 amount) public view {
```
avniculae, line 81: This is identical to `testPreviewSwapFromGHO` above, and `testPreviewSwapFromSGHOStataUSDX_4626_TOKEN` is likewise identical to `testPreviewSwapFromSGHOUSDX_4626_TOKEN`. Looks like one of each pair was meant to use `USDX_TOKEN` — as is, previews with the underlying token in/out are never tested.

tests/unit/GhoRouter/TestGhoRouterSwap.t.sol
```diff
@@ -0,0 +1,898 @@
...
+      address(GHO_GSM_4626),
+      amount,
+      0,
+      USER,
+      block.timestamp
+    );
+    assertEq(outputAmount, expectedAmount, 'Should receive output token');
+  }
+
+  function testSwapFromGHOToUSDX_4626_TOKEN(uint256 amount) public {
```
avniculae: Both FromGHO success tests output the stata token; there's no GHO -> `USDX_TOKEN` (underlying) test, so the unwrap-on-buy path is only exercised from sGHO.

src/contracts/misc/interfaces/IGhoRouter.sol
```diff
@@ -0,0 +1,189 @@
...
+   * @notice Emitted when a swap is completed
+   * @param caller The address of the initiator of the swap
+   * @param tokenIn The address of the token to swap from
+   * @param tokenOut The address of the token to swap to
+   * @param exactAmountIn The amount of tokenIn swapped
+   * @param amountOut The amount of tokenOut received
+   * @param recipient The address of the recipient of tokenOut
+   */
+  event Swap(
+    address caller,
```
avniculae: All three indexed slots are spent on `tokenIn`/`tokenOut`/`recipient`; `caller` is arguably the field integrators/monitoring will filter on most. Worth reconsidering which three get indexed.

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,567 @@
...
+    uint256 minAmountOut,
+    address recipient,
+    uint256 deadline
+  ) external returns (uint256) {
+    _validateInputs(exactAmountIn, recipient, deadline);
+
+    uint256 amountOut;
+    if (tokenIn == GHO) {
+      if (tokenOut == sGHO) {
+        amountOut = _ghoToSGho(exactAmountIn, minAmountOut, recipient);
```
avniculae, line 65: The GHO <-> sGHO paths ignore `gsm` entirely, so an integrator passing a wrong GSM here still succeeds — worth a note in the natspec/docs. Also, GHO -> GHO falls into `_ghoToToken` and reverts `InvalidToken` rather than `InvalidTokenPair`.

tests/mocks/MockBadGsm.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.27;
+
+contract BadGsm {
```
avniculae: nit: other mocks name the contract `Mock*` to match the file; `BadGsm`/`BadStata` inside `MockBad*.sol` files break that.

src/contracts/misc/GhoRouter.sol
avniculae: +1, this should be enforced: the router pulls whatever `getAssetAmountForBuyAsset` returns, so with a nonstandard GSM and a larger router allowance the user can be charged more than `exactAmountIn`. A `require(ghoUsed <= exactAmountIn)` here closes it.

tests/unit/GhoRouter/TestGhoRouterSetTokenToGsm.t.sol
avniculae: Only the token -> GHO direction was added (`testSwapToGHORegularGsm`); the GHO -> token, sGHO -> token and token -> sGHO flows still have no plain-GSM tests.

src/contracts/misc/GhoRouter.sol
avniculae: The disclaimers were dropped from the router helpers, but `IGhoRouter.previewSwap` still carries "This is an estimation..." and the router uses `@inheritdoc`, so it survives — should be removed from the interface too.

src/contracts/misc/GhoRouter.sol
avniculae: Adding to this: `sellAsset` can pull less than the approved amount (its calc rounds down), so a live residual approval to the GSM remains, and `revokeGsm` doesn't clear approvals — a revoked GSM keeps pull rights over any dust left in the router. Dust-scale, but `forceApprove(0)` after the calls closes it.

src/contracts/misc/GhoRouter.sol
avniculae: Correction to the above: one positional call remains — the stata deposit in `_sellTokenForGho` (GhoRouter.sol L450, `IERC4626(gsmAsset).deposit(exactAmountIn, address(this))`). The rest of the deposit/redeem/sellAsset/buyAsset calls are named.

src/contracts/misc/GhoRouter.sol
avniculae: Correction to the above: the stata deposit in `_sellTokenForGho` (GhoRouter.sol L450) is still positional; all the other similar calls are named.

src/contracts/misc/GhoRouter.sol
avniculae: Correction to the above: the code-level fix stands, but the natspec still equates `gsmAsset` with "static aToken" (L433, L461-466, L540 "the static aToken the GSM settles in", and the repeated "underlying token or static aToken" param docs), which is inaccurate for plain GSMs — the terminology point is still open.

src/contracts/misc/GhoRouter.sol
efecarranza: done

src/contracts/misc/interfaces/IGhoRouter.sol
Kogaroshi: This was modified to drop `recipient` from the indexed fields in favor of `caller`. As stated by Alex, `caller` is an important filter, and I think the assets (`tokenIn` & `tokenOut`) are also necessary to have efficient filters there, so this would be the best possible choice here.

src/contracts/misc/GhoRouter.sol
Kogaroshi: Natspec was updated, no "GSM" or "static aToken" mentions left at the listed lines. Closing this.

src/contracts/misc/GhoRouter.sol
Kogaroshi: missing change in L450 was made, closing this.

src/contracts/misc/GhoRouter.sol
Kogaroshi: see https://github.com/aave-dao/gho-origin/pull/34#discussion_r3569314716

src/contracts/misc/GhoRouter.sol
Kogaroshi: line was remove from Interface Natspec

src/contracts/misc/script.ts
```diff
@@ -0,0 +1,62 @@
+import Safe from '@safe-global/protocol-kit';
```
Kogaroshi: why is this in src/ ?

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,569 @@
...
+  /// @inheritdoc IGhoRouter
+  function swap(
+    address tokenIn,
+    address tokenOut,
+    address gsm,
+    uint256 exactAmountIn,
+    uint256 minAmountOut,
+    address recipient,
+    uint256 deadline
+  ) external returns (uint256) {
```
Kogaroshi: let's add the Transient Reentrancy Guard, as a precaution : we cannot predict future GSM behaviors

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,569 @@
...
+   * @dev Redeems sGHO into GHO, then swaps through GSM into a caller-selected token.
+   * @dev An almost negligible amount of dust can be left unsued, would cost more in gas than amount returned.
+   * @param token Output token (either GSM static aToken or aToken's underlying).
+   * @param gsm Whitelisted GSM used for the swap.
+   * @param exactAmountIn Amount of sGHO shares pulled from the caller.
+   * @param minOutputAmount Minimum acceptable output-token amount.
+   * @param recipient Address that receives the resulting token.
+   * @return outputAmount Amount of output tokens sent to `recipient`.
+   */
+  function _sGhoToToken(
```
Kogaroshi, line 334: Let's add a `RequiredGhoGreaterThanExpectedAmount` revert here as well

src/contracts/misc/script.ts
efecarranza: oops, removed! was upgrading a SAFE and i guess just ran it from the terminal i had open

src/contracts/misc/GhoRouter.sol
efecarranza: @Kogaroshi i need to bump the evm version for that to work, is that ok?

Kogaroshi: Let's not touch compiler version for now, we can use the traditional ReentrancyGuard then

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,572 @@
...
+
+    emit Swap(msg.sender, tokenIn, tokenOut, exactAmountIn, amountOut, recipient);
+
+    return amountOut;
+  }
+
+  /// @inheritdoc IGhoRouter
+  function allowGsm(address gsm) external onlyOwner {
+    require(gsm != address(0), ZeroAddress());
+    require(allowedGsm[gsm] == false, GsmAlreadySet());
```
Kogaroshi: nit: could be `require(!allowedGsm[gsm], GsmAlreadySet());` for consistency with L108

src/contracts/misc/GhoRouter.sol
Kogaroshi: @efecarranza can you please consider changing the `Ownable` there, imo it's okay to break convention from this repo if it is to increase security.

src/contracts/misc/GhoRouter.sol
```diff
@@ -0,0 +1,572 @@
...
+    uint256 exactAmountIn,
+    uint256 amountToBuy,
+    address outputReceiver,
+    uint256 minOutputAmount
+  ) internal returns (uint256) {
+    IERC20(GHO).forceApprove(gsm, exactAmountIn);
+    (uint256 underlyingAmount, ) = IGsm(gsm).buyAsset({
+      minAmount: amountToBuy,
+      receiver: address(this)
+    });
```
Kogaroshi, line 501: let's also add `IERC20(GHO).forceApprove(gsm, 0);` here for consistency, or drop from both sides for gas savings & consistency, or we can consider granting MAX approval to the GSMs for GHO & their `gsmAsset` if gas costs reduction is the priority.
@avniculae wdyt ?

Kogaroshi (approved):

PR #40, opened 2026-05-28
feat: add StkGhoMigrator & tests

AGMASO:
## Summary

Adds `StkGhoMigrator`, a helper contract that migrates a user's full stkGHO position into the sGHO ERC4626 vault in a single transaction (cooldown + redeem + deposit batched).

## Changes

- `src/contracts/misc/StkGhoMigrator.sol` + interfaces under `src/contracts/misc/interfaces/`.
- Fork and stateless fuzz tests under `tests/misc/` (auto-skipped if `RPC_MAINNET` is not set).
- `docs/stkgho-migrator.md` with flow, role handshake and testing notes.

## Prerequisite

The migrator must hold the stkGHO `CLAIM_HELPER` role (id `2`): current admin calls `setPendingAdmin(2, migrator)` on stkGHO, then anyone calls `claimHelperRole()` on the migrator.


src/contracts/misc/interfaces/IStakeToken.sol
```diff
@@ -0,0 +1,213 @@
+// SPDX-License-Identifier: BUSL-1.1
```
Kogaroshi, line 1: we have aave-v3-origin dependency in this repo, we can import from there instead of copying this interface

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,104 @@
...
+    _pause();
+  }
+
+  /// @inheritdoc IStkGhoMigrator
+  function unpause() external onlyOwnerOrPauseGuardian {
+    _unpause();
+  }
+
+  /// @inheritdoc IStkGhoMigrator
+  function rescue(address token, address to, uint256 amount) external onlyOwner {
```
Kogaroshi: move this before pause/unpause, to respect permission modifiers order

tests/misc/interfaces/IRoleManager.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
```
Kogaroshi: same, I think we could be able to import this from aave-v3 ?

tests/misc/StkGhoMigratorForkTest.t.sol
```diff
@@ -0,0 +1,368 @@
...
+import {Ownable} from 'openzeppelin-contracts/contracts/access/Ownable.sol';
+import {Pausable} from 'openzeppelin-contracts/contracts/utils/Pausable.sol';
+import {IERC20} from 'openzeppelin-contracts/contracts/token/ERC20/IERC20.sol';
+import {IERC4626} from 'openzeppelin-contracts/contracts/interfaces/IERC4626.sol';
+import {IRoleManager} from './interfaces/IRoleManager.sol';
+
+contract StkGhoMigratorForkTest is Test {
+  StkGhoMigrator public migrator;
+  address public user = makeAddr('USER');
+  address public user2 = makeAddr('USER2');
```
Kogaroshi: given how we mostly use `user2` in tests, we might want to rename `user2` to `user`, and the current `user` to `externalUser` or `invalidUser` ?

src/contracts/misc/interfaces/IStakeToken.sol
AGMASO: I checked the interfaces included in `aave-v3-origin` and i can't find the ones i need. There is only a slim interface available, but it doesn’t include all the methods I need. So I don’t have any other option but to include them locally, as I’m currently doing.

tests/misc/interfaces/IRoleManager.sol
AGMASO: I will delete this one and merge it into `IStakeToken`, since it is actually inherited there.

src/contracts/misc/interfaces/IStakeToken.sol
Kogaroshi: mb, it's not in aave-v3-origin, but in https://github.com/aave-dao/aave-address-book/blob/97f2917aa593237e7fb8d91c9b7879bf471068c9/src/common/IStakeToken.sol

AGMASO: Gotcha! I’ll install the library then and remove the local version.

AGMASO: done @Kogaroshi

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,104 @@
...
+import {IStkGhoMigrator} from 'src/contracts/misc/interfaces/IStkGhoMigrator.sol';
+import {IStakeToken} from 'aave-address-book/common/IStakeToken.sol';
+
+/**
+ * @title StkGhoMigrator
+ * @author Aave Labs
+ * @notice Migrates stkGHO positions into the sGHO ERC4626 vault in a single transaction.
+ * @dev The contract must hold the stkGHO claim helper role to call cooldownOnBehalfOf and redeemOnBehalf.
+ */
+contract StkGhoMigrator is IStkGhoMigrator, Pausable, Ownable {
```
avniculae: do we want to make it Ownable2Step?

avniculae: also given we use guardian, let's use OwnableWithGuardian from solidity-utils, wdyt?

https://github.com/aave-dao/solidity-utils/blob/main/src/contracts/access-control/OwnableWithGuardian.sol

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /// @inheritdoc IStkGhoMigrator
+  function pause() external onlyOwnerOrPauseGuardian {
+    _pause();
+  }
+
+  /// @inheritdoc IStkGhoMigrator
+  function unpause() external onlyOwnerOrPauseGuardian {
+    _unpause();
```
avniculae, line 68: should only owner unpause?

Kogaroshi: might be a good compromise yeah, to have the pause as a quick trigger response, and unpause only after correctly reviewing.

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    STKGHO.cooldownOnBehalfOf(msg.sender);
+    STKGHO.redeemOnBehalf(msg.sender, address(this), stkGhoShares);
+    uint256 ghoRedeemed = GHO.balanceOf(address(this)) - ghoBalanceBefore;
+    require(ghoRedeemed == stkGhoShares, UnexpectedGhoRedeemed());
+
+    uint256 sghoSharesReceived = SGHO.deposit(ghoRedeemed, msg.sender);
+    require(sghoSharesReceived != 0, NoSGhoSharesReceived());
+
+    emit MigrationBatchExecuted(msg.sender, stkGhoShares, ghoRedeemed, sghoSharesReceived);
```
avniculae: I wouldn't emit both stkGhoShares and ghoRedeemed, we require above they are equal

tests/misc/StkGhoMigratorForkTest.t.sol
```diff
@@ -0,0 +1,366 @@
...
+    vm.deal(user, 1 ether);
+    deal(address(GHO), user, 1_000e18);
+    vm.startPrank(user);
+    IERC20(GHO).approve(address(STKGHO), 90e18);
+    STKGHO.stake(user, 90e18);
+    vm.warp(block.timestamp + 30 days);
+    vm.stopPrank();
+    _;
+  }
+  modifier changeCooldownToNotZero() {
```
avniculae, line 23: nit: new line between modifiers

remappings.txt
```diff
@@ -2,3 +2,4 @@ forge-std/=lib/forge-std/src/
 aave-v3-origin/=lib/aave-v3-origin/src/
 aave-v3-origin-tests/=lib/aave-v3-origin/tests/
 solidity-utils/=lib/aave-v3-origin/lib/solidity-utils/src/
+aave-address-book/=lib/aave-address-book/src/
```
avniculae: I wouldn't install address book here, gho-deploy uses gho-origin and also imports address-book. I think we can simply define the IStakeToken interface, wdyt? cc @Kogaroshi

Kogaroshi: I missed that part with gho-deploy, so yeah maybe better to revert to copy and pasting the interface here then.

AGMASO: We had it like that before, but in the last round of reviews I was asked to change it to use the import and delete `IStakeToken`. Should we revert back to the previous approach?

src/contracts/misc/StkGhoMigrator.sol
AGMASO: what should we do here? should we inherit both Contracts? or just Ownable2Step?

AGMASO: Heads up on one behavior difference: the inherited updateGuardian is onlyOwnerOrGuardian and performs no zero-address or same-value checks, whereas our current setPauseGuardian is onlyOwner and validates both. To preserve that stricter behavior we'd need to override updateGuardian, which brings back some of the boilerplate the lib was supposed to remove. Maybe we should add only Ownable2Step?

miguelmtzinf: Let's use `Onwable2StepWithGuardian`

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,104 @@
...
+  /// @inheritdoc IStkGhoMigrator
+  function unpause() external onlyOwner {
+    _unpause();
+  }
+
+  /// @inheritdoc IStkGhoMigrator
+  function migrate() external whenNotPaused {
+    require(STKGHO.getCooldownSeconds() == 0, CooldownPeriodNotZero());
+
+    uint256 stkGhoShares = STKGHO.balanceOf(msg.sender);
```
miguelmtzinf:
```suggestion
    uint256 amountToRedeem = STKGHO.balanceOf(msg.sender);
```

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    STKGHO.cooldownOnBehalfOf(msg.sender);
+    STKGHO.redeemOnBehalf(msg.sender, address(this), stkGhoShares);
+    uint256 ghoRedeemed = GHO.balanceOf(address(this)) - ghoBalanceBefore;
+    require(ghoRedeemed == stkGhoShares, UnexpectedGhoRedeemed());
+
+    uint256 sghoSharesReceived = SGHO.deposit(ghoRedeemed, msg.sender);
+    require(sghoSharesReceived != 0, NoSGhoSharesReceived());
+
+    emit MigrationBatchExecuted(msg.sender, ghoRedeemed, sghoSharesReceived);
```
miguelmtzinf: there is no batch indeed. We can rename to `stkGhoMigrated`. I think emitting the gho amounted is enough tho.  I d do {user, amount} instead too

src/contracts/misc/interfaces/IStkGhoMigrator.sol
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * @title IStkGhoMigrator
+ * @notice Interface for migrating stkGHO positions into the sGHO ERC4626 vault in a single transaction.
+ */
+interface IStkGhoMigrator {
+  /// @notice Emitted when a migration is executed.
+  /// @param user The address of the user executing the migration.
+  /// @param amount The amount of stkGHO migrated.
+  event StkGhoMigrated(address indexed user, uint256 amount);
```
avniculae, line 16: let's emit 2 amounts here: stkGho removed, and sGho received, wdyt?

src/contracts/misc/interfaces/IStkGhoMigrator.sol
```diff
@@ -0,0 +1,89 @@
...
+
+  /// @notice Thrown when the stkGHO cooldown period is not zero.
+  error CooldownPeriodNotZero();
+  /// @notice Thrown when the user has no stkGHO shares to redeem.
+  error NoStkGhoSharesToRedeem();
+  /// @notice Thrown when the redeemed shares did not return the expected amount of GHO.
+  error UnexpectedGhoRedeemed();
+  /// @notice Thrown when an input address is zero.
+  error InvalidAddressZero();
```
avniculae: We can remove "Zero", just "InvalidAddress"

src/contracts/misc/interfaces/IStkGhoMigrator.sol
```diff
@@ -0,0 +1,89 @@
...
+  error NoSGhoSharesReceived();
+
+  /**
+   * @notice Claims the stkGHO claim helper role.
+   * @dev This contract must have been set as the pending admin for the claim helper role.
+   */
+  function claimHelperRole() external;
+
+  /**
+   * @notice Sets the pending admin for the stkGHO claim helper role.
```
avniculae, line 39: Let's document this is only owner, similar to how we document in rescue

src/contracts/misc/interfaces/IStkGhoMigrator.sol
```diff
@@ -0,0 +1,89 @@
...
+   * @dev Only callable by the owner.
+   * @param token The ERC20 token to rescue.
+   * @param to The address that will receive the rescued tokens.
+   * @param amount The amount of tokens to rescue.
+   */
+  function rescue(address token, address to, uint256 amount) external;
+
+  /**
+   * @notice Pauses migrations.
+   * @dev Callable by the owner or the pause guardian.
```
avniculae:
```suggestion
   * @dev Only callable by the owner or the pause guardian.
```

src/contracts/misc/interfaces/IStkGhoMigrator.sol
```diff
@@ -0,0 +1,89 @@
...
+   */
+  function pause() external;
+
+  /**
+   * @notice Unpauses migrations.
+   * @dev Callable by the owner only.
```
avniculae:
```suggestion
   * @dev Only callable by the owner only.
```

src/contracts/misc/interfaces/IStkGhoMigrator.sol
```diff
@@ -0,0 +1,89 @@
...
+   */
+  function unpause() external;
+
+  /**
+   * @notice Migrates the caller's full stkGHO position into the sGHO ERC4626 vault.
+   * @dev Reverts when the contract is paused.
```
avniculae: Let's expand here a bit, it can revert also when cooldown period is not 0, or when contract does not hold the claimer role, or when stkGho price share is not 1:1, or when sGho received is 0, or when stkgho balance is 0. I think we can explain a bit what are the operations performed in this action.

src/contracts/misc/interfaces/IStkGhoMigrator.sol
```diff
@@ -0,0 +1,89 @@
...
+   */
+  function migrate() external;
+
+  /**
+   * @notice Returns the stkGHO token contract.
+   */
+  function STKGHO() external view returns (IStakeToken);
```
avniculae, line 80: I think address type should be address in all of these getters

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,103 @@
...
+ * @title StkGhoMigrator
+ * @author Aave Labs
+ * @notice Migrates stkGHO positions into the sGHO ERC4626 vault in a single transaction.
+ * @dev The contract must hold the stkGHO claim helper role to call cooldownOnBehalfOf and redeemOnBehalf.
+ */
+contract StkGhoMigrator is IStkGhoMigrator, Pausable, Ownable2Step, OwnableWithGuardian {
+  using SafeERC20 for IERC20;
+
+  /// @notice The stkGHO token contract
+  IStakeToken public constant STKGHO = IStakeToken(0x1a88Df1cFe15Af22B3c4c783D4e6F7F9e0C1885d);
```
avniculae, line 22: we can import these from address-book, we have it as dependency, instead of hardcoding. we can add tests where we compare these to hardcoded values

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,103 @@
...
+  /// @notice The stkGHO claim helper role ID
+  uint256 public constant CLAIM_HELPER_ROLE = 2;
+
+  /**
+   * @notice Initializes the contract owner and grants the sGHO vault maximum GHO allowance.
+   * @param initialOwner The address of the contract owner.
+   * @param initialPauseGuardian The address of the initial pause guardian.
+   */
+  constructor(
+    address initialOwner,
```
avniculae: nit: initialOwner_, initialPauseGuardian_

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,103 @@
...
+  IStakeToken public constant STKGHO = IStakeToken(0x1a88Df1cFe15Af22B3c4c783D4e6F7F9e0C1885d);
+  /// @notice The sGHO ERC4626 vault contract
+  IERC4626 public constant SGHO = IERC4626(0xE1753F2e00940cC31213dd92013cF019DFE4ca1d);
+  /// @notice The GHO token contract
+  IERC20 public constant GHO = IERC20(0x40D16FC0246aD3160Ccc09B8D0D3A2cD28aE6C2f);
+  /// @notice The stkGHO claim helper role ID
+  uint256 public constant CLAIM_HELPER_ROLE = 2;
+
+  /**
+   * @notice Initializes the contract owner and grants the sGHO vault maximum GHO allowance.
```
avniculae, line 31:
```suggestion
   * @notice Constructor
   * @dev Initializes the contract owner and grants the sGHO vault maximum GHO allowance.
```

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,103 @@
...
+  constructor(
+    address initialOwner,
+    address initialPauseGuardian
+  ) OwnableWithGuardian(initialOwner, initialPauseGuardian) {
+    require(initialPauseGuardian != address(0), InvalidAddressZero());
+    GHO.forceApprove(address(SGHO), type(uint256).max);
+  }
+
+  /// @inheritdoc IStkGhoMigrator
+  function claimHelperRole() external {
```
avniculae: shoul this be `whenNotPaused` ?

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,103 @@
...
+import {IStkGhoMigrator} from 'src/contracts/misc/interfaces/IStkGhoMigrator.sol';
+import {IStakeToken} from 'src/contracts/misc/interfaces/IStakeToken.sol';
+
+/**
+ * @title StkGhoMigrator
+ * @author Aave Labs
+ * @notice Migrates stkGHO positions into the sGHO ERC4626 vault in a single transaction.
+ * @dev The contract must hold the stkGHO claim helper role to call cooldownOnBehalfOf and redeemOnBehalf.
+ */
+contract StkGhoMigrator is IStkGhoMigrator, Pausable, Ownable2Step, OwnableWithGuardian {
```
avniculae:
```suggestion
contract StkGhoMigrator is IStkGhoMigrator, Pausable, Ownable2StepWithGuardian {
```

we can also overrides we have for duplicate fns

tests/misc/StkGhoMigratorForkTest.t.sol
```diff
@@ -0,0 +1,385 @@
...
+  address public ownerMigrator = makeAddr('OWNER_MIGRATOR');
+  address public pauseGuardian = makeAddr('PAUSE_GUARDIAN');
+  IStakeToken public constant STKGHO = IStakeToken(0x1a88Df1cFe15Af22B3c4c783D4e6F7F9e0C1885d);
+  IERC4626 public constant SGHO = IERC4626(0xE1753F2e00940cC31213dd92013cF019DFE4ca1d);
+  IERC20 public constant GHO = IERC20(0x40D16FC0246aD3160Ccc09B8D0D3A2cD28aE6C2f);
+  uint256 constant CLAIM_HELPER_ROLE = 2;
+  uint256 constant COOLDOWN_ADMIN_ROLE = 1;
+
+  modifier depositStkGhoReadyToRedeem() {
+    vm.deal(user, 1 ether);
```
avniculae, line 14: what's this needed for?

tests/misc/StkGhoMigratorForkTest.t.sol
```diff
@@ -0,0 +1,385 @@
...
+  // --- Tests migrate ---
+  function test_Migrate() public depositStkGhoReadyToRedeem {
+    uint256 stkGhoShares = STKGHO.balanceOf(user);
+    uint256 sGhoSharesBefore = SGHO.balanceOf(user);
+
+    vm.prank(user);
+    migrator.migrate();
+
+    assertEq(STKGHO.balanceOf(user), 0);
+    assertGt(SGHO.balanceOf(user), sGhoSharesBefore);
```
avniculae: let's be a bit more tight here, this check is loose

tests/misc/StkGhoMigratorForkTest.t.sol
```diff
@@ -0,0 +1,385 @@
...
+    vm.prank(adminCooldown);
+    STKGHO.setPendingAdmin(COOLDOWN_ADMIN_ROLE, address(this));
+    STKGHO.claimRoleAdmin(COOLDOWN_ADMIN_ROLE);
+    STKGHO.setCooldownSeconds(86400); // Set cooldown to 1 day
+    _;
+  }
+
+  function setUp() public {
+    // Skip if RPC_MAINNET env variable is not set.
+    string memory rpc = vm.envOr('RPC_MAINNET', string(''));
```
avniculae, line 34: We should have a test base that can run with no fork, and then we can inherit it here for fork tests. this way we execute same tests and we run them in cicd as well. (ofc we need different setUp() in the unit tests)

src/contracts/misc/interfaces/IStkGhoMigrator.sol
AGMASO: It was like that before, but I got a comment from @miguelmtzinf asking me to change it to its current behavior.

tests/misc/StkGhoMigratorForkTest.t.sol
AGMASO: `COOLDOWN_ADMIN_ROLE` is used by the cooldown revert test to impersonate the stkGHO cooldown admin and set a non-zero cooldown.

The modifier prepares a redeemable stkGHO position for the migration tests.

src/contracts/misc/interfaces/IStkGhoMigrator.sol
AGMASO: Changing these getters to return `address` breaks the public state variable overrides, because the constants are currently typed as `IStakeToken`, `IERC4626`, and `IERC20`.

To make the interface return `address`, we would also need to change the constants to `address` and cast at every usage site. In my opinion that adds noise and reduces readability.

Do you consider keeping the typed getters a bad practice for production?

tests/misc/StkGhoMigratorForkTest.t.sol
AGMASO: I moved the tests that are environment-agnostic into a shared base so they run both in the no-fork unit suite and in the fork suite through inheritance.

I kept some tests unit-only where the goal is to validate the migrator behavior with mocked external calls, and kept the integration-heavy cases fork-only where they depend on real stkGHO/sGHO/GHO behavior.

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * @notice Initializes the contract owner and grants the sGHO vault maximum GHO allowance.
+   * @param initialOwner_ The address of the contract owner.
+   * @param initialPauseGuardian_ The address of the initial pause guardian.
+   */
+  constructor(
+    address initialOwner_,
+    address initialPauseGuardian_
+  ) Ownable2StepWithGuardian(initialOwner_, initialPauseGuardian_) {
+    require(initialPauseGuardian_ != address(0), InvalidAddress());
```
avniculae, line 39: uoff, we need to fix Ownable2StepWithGuardian to check this, code was left the same as non 2step version

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,90 @@
...
+  function migrate() external whenNotPaused {
+    require(STKGHO.getCooldownSeconds() == 0, CooldownPeriodNotZero());
+
+    uint256 amountToRedeem = STKGHO.balanceOf(msg.sender);
+    require(amountToRedeem != 0, NoStkGhoSharesToRedeem());
+
+    uint256 ghoBalanceBefore = GHO.balanceOf(address(this));
+
+    STKGHO.cooldownOnBehalfOf(msg.sender);
+    STKGHO.redeemOnBehalf(msg.sender, address(this), amountToRedeem);
```
avniculae: let's use named args here pls

src/contracts/misc/StkGhoMigrator.sol
avniculae: tracked here https://github.com/aave-dao/solidity-utils/issues/6

src/contracts/misc/StkGhoMigrator.sol
AGMASO: done

src/contracts/misc/StkGhoMigrator.sol
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.27;
+
+import {Ownable2StepWithGuardian} from 'solidity-utils/contracts/access-control/Ownable2StepWithGuardian.sol';
```
avniculae, line 4: I think we are missing a stk gho migrator deployment procedure, to be able to use it in gho-deploy repo.

AGMASO: merged

avniculae (approved):

PR #41, opened 2026-05-28
refactor: Upgrade to OZ v5.1 (excl. GhoToken & UpgradeableGhoToken)

avniculae:


src/contracts/gho/interfaces/IGhoToken.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import {IERC20} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol';
+import {IERC20} from 'openzeppelin-contracts/contracts/token/ERC20/IERC20.sol';
```
avniculae: I changed this bc it is the same (apart from some name changes), but left IAccessControl as before. AccessControl has bytecode changes in v5.1 vs v4.7. wdyt?

src/contracts/facilitators/gsm/Gsm.sol
```diff
@@ -4,10 +4,10 @@ pragma solidity ^0.8.10;
...
 import {IERC20} from 'aave-v3-origin/contracts/dependencies/openzeppelin/contracts/IERC20.sol';
 import {GPv2SafeERC20} from 'aave-v3-origin/contracts/dependencies/gnosis/contracts/GPv2SafeERC20.sol';
-import {EIP712} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/utils/cryptography/EIP712.sol';
-import {SignatureChecker} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/utils/cryptography/SignatureChecker.sol';
-import {SafeCast} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/utils/math/SafeCast.sol';
-import {AccessControl} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/access/AccessControl.sol';
+import {EIP712} from 'openzeppelin-contracts/contracts/utils/cryptography/EIP712.sol';
+import {SignatureChecker} from 'openzeppelin-contracts/contracts/utils/cryptography/SignatureChecker.sol';
+import {SafeCast} from 'openzeppelin-contracts/contracts/utils/math/SafeCast.sol';
+import {AccessControl} from 'openzeppelin-contracts/contracts/access/AccessControl.sol';
```
avniculae, line 10: Bc we use new AccessControl, existing Gsms cannot be upgraded with new implementation since storage layout is different (see diff in test as well). I think this is fine as we discussed about this in the past, but I wanted to flag and double check, wdyt?

src/contracts/gho/UpgradeableERC20.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT-only
 pragma solidity ^0.8.0;
 
-import {IERC20} from 'src/contracts/dependencies/openzeppelin-contracts/contracts/token/ERC20/IERC20.sol';
+import {IERC20} from 'openzeppelin-contracts/contracts/token/ERC20/IERC20.sol';
 
 /**
  * @title UpgradeableERC20
```
avniculae, line 7: I'm wondering if, given these changes in this PR, we should change `GhoTokenProcedure` to use transparent proxy from v5.1. we can adjsut tests, and then we don't need to override it anymore in gho-deploy, I think this is cleaner. wdyt? 

Note, this will not change the bytyecode of the impl, just the one of the proxy (which anyways is not used, since we override it in gho-deploy)

src/contracts/facilitators/gsm/Gsm.sol
efecarranza: in the past, we've migrated GSMs without upgrading but just deploying and seizing the previuos ones. i don't mind this change and i thin it's better to update the dependencies and migrate than not update dependencies just for an upgrade so i'd make the changes

src/contracts/gho/interfaces/IGhoToken.sol
efecarranza: if we are wanting to keep GhoToken the same in terms of bytecode that makes sense. is there a specific reason why we always want to keep the bytecode the same even for newer deployments?

src/contracts/gho/UpgradeableERC20.sol
efecarranza: i'd go ahead and update this one too myself, and move to 5.1. i don't know all the details but there might be fixes from 5.1 to 4.8.1 too

tests/unit/TestGhoAaveSteward.t.sol
```diff
@@ -814,4 +814,31 @@ contract TestGhoAaveSteward is TestGhoBase, GhoStewardProcedure {
         ) / 1e23
       ); // Convert to bps
   }
+
+  function testConstructorEmitsOwnershipTransferred() public {
+    address expected = vm.computeCreateAddress(address(this), vm.getNonce(address(this)));
+    vm.expectEmit(expected);
```
efecarranza: I like doing vm.expectemit(true,true,true,true, expected) to ensure the emitter and data are correct

src/contracts/gho/interfaces/IGhoToken.sol
miguelmtzinf: For maintenance purposes. Having different versions of the code across chains is error-prone

miguelmtzinf: @avniculae i d revert, so it's clear token uses local deps always (no exceptions).

tests/helpers/ErrorsLib.sol
```diff
@@ -19,14 +19,3 @@ library AccessControlErrorsLib {
     // Excludes contract from coverage.
   }
 }
-
-library OwnableErrorsLib {
```
miguelmtzinf, line 23: if we are rming this one, can we remove the entire file also?

avniculae: I wanted to keep the other helper bc gho token still uses old access control, I don't want to inline it in ~10 places

Kogaroshi (approved):

efecarranza (approved):

PR #42, opened 2026-06-30
feat(sGho): linear fixed-APR yield, storage repack & mainnet patch

avniculae:
## Summary

Reworks sGHO yield accrual to be **linear (fixed APR)** instead of compounding, and repacks the namespaced storage now that `ratePerSecond` is no longer needed.

## Yield model
- `_getCurrentYieldIndex` accrues linearly: `index += targetRate * timeElapsed / year`. Dividing by the year last keeps a full APR period exact, and it never compounds.
- The index is checkpointed to storage only when the rate changes (`setTargetRate`), not on every deposit/withdraw/transfer. This removes the per-action `_updateYieldIndex` and the rounding/DoS surface from frequent tiny updates.

## Storage
- Removed `ratePerSecond`. Repacked `sGhoStorage` into a single slot: `uint120 yieldIndex`, `uint40 lastUpdate`, `uint16 targetRate`, `uint40 supplyCap`.
- `supplyCap` is now stored in whole GHO units (no decimals), scaled by `10 ** decimals()` in `maxDeposit`.
- Getters (`yieldIndex`/`lastUpdate`/`supplyCap`) return `uint256` and `setSupplyCap` takes `uint256`, aligned with aave-v4 (which exposes `uint256` scalar getters over packed storage). `setSupplyCap` now takes whole GHO units.

## Deployment & mainnet migration
- `sGhoInstance` — standard initializer for fresh deploys (revision 2, aligned across chains).
- `sGhoInstanceStoragePatch` — one-off implementation for the live Ethereum mainnet vault. Its `initialize()` (reinitializer 2) self-reads the previous two-slot layout, rewrites the repacked slot, converts `supplyCap` to whole units, and wipes the orphaned second slot. Run atomically via `upgradeToAndCall`; governance should `setTargetRate(currentRate)` first so the index is checkpointed and the exchange rate stays continuous across the upgrade.

## Gas
The first commit adds gas snapshots on the unchanged contract; the change commit's snapshot diff shows the effect — every common op is cheaper (`transfer` −9.9k, `deposit`/`withdraw`/`redeem` ~−7.6k, `setTargetRate` −5.4k) because `_update` no longer checkpoints.

## Tests
- Updated unit/precision suites for the linear model and the new layout.
- Added mainnet fork tests (`tests/fork/`, gated on `RPC_MAINNET`): every affected storage field (`yieldIndex`/`lastUpdate`/`targetRate`/`supplyCap`) is read before and after the upgrade and is unchanged (supplyCap converted to whole units); the old second slot and the unused high bits of slot 0 are zeroed; an existing holder's position is preserved and they can still redeem; existing and new positions accrue linearly post-upgrade.

## Notes
- Removed `docs/sgho-precision-analysis` — it analyzed the previous compounding model and no longer applies.


src/contracts/sgho/sGho.sol
```diff
@@ -36,24 +39,20 @@ contract sGho is
...
+    uint40 supplyCap; // 5 bytes - maximum total assets allowed in the vault, in whole GHO units
   }
 
   // keccak256(abi.encode(uint256(keccak256("gho.storage.sGho")) - 1)) & ~bytes32(uint256(0xff))
-  bytes32 private constant sGhoStorageLocation =
+  bytes32 internal constant sGhoStorageLocation =
     0x52190d4bcaca04cac5a7c2ae78ea3854d285be3b91819fb1b3ed9862d9a9a400;
 
-  function _getSGhoStorage() private pure returns (sGhoStorage storage $) {
+  function _getSGhoStorage() internal pure returns (sGhoStorage storage $) {
```
Kogaroshi, line 55: I think here is the only place with 'SGho' instead of 'sGho' ? maybe we can rename for consistency ?

Kogaroshi: nvm, it was already named like that before, let's not change then

tests/fork/sGhoInstanceStoragePatch.fork.t.sol
```diff
@@ -0,0 +1,182 @@
...
+    _upgrade();
+
+    // The holder's position is consistent across the upgrade
+    assertEq(sgho.balanceOf(HOLDER), sharesBefore, 'shares changed');
+    assertApproxEqAbs(sgho.convertToAssets(sharesBefore), assetsBefore, 1, 'asset value changed');
+    assertEq(sgho.totalSupply(), totalSupplyBefore, 'totalSupply changed');
+
+    // Migrated fields are correct in the new layout
+    assertEq(sgho.targetRate(), rateBefore, 'targetRate not preserved');
+    assertEq(sgho.supplyCap(), 400_000_000, 'supplyCap not converted to whole units');
```
Kogaroshi: can we replace with fetched `supplyCap` pre upgrade instead ? Cap might change before we do the update

Kogaroshi (approved):
small nit, otherwise lgtm on the logic change & new index calculations

src/contracts/sgho/instances/sGhoInstanceStoragePatch.sol
```diff
@@ -0,0 +1,58 @@
...
+  function initialize() external reinitializer(SGHO_REVISION) {
+    uint256 slot0;
+    uint256 slot1;
+    assembly {
+      slot0 := sload(sGhoStorageLocation)
+      slot1 := sload(add(sGhoStorageLocation, 1))
+    }
+
+    uint120 yieldIndex_ = uint256(uint176(slot0)).toUint120();
+    uint40 lastUpdate_ = uint40(slot0 >> 176);
```
avniculae: I think we should revert if lastUpdate_ is not block.timestamp

src/contracts/sgho/instances/sGhoInstance.sol
```diff
@@ -0,0 +1,32 @@
...
+import {sGho} from 'src/contracts/sgho/sGho.sol';
+
+/**
+ * @title sGhoInstance
+ * @author kpk, TokenLogic & Aave Labs
+ * @notice Deployable implementation of the sGHO vault.
+ */
+contract sGhoInstance is sGho {
+  /// @dev Aligned with the mainnet storage-patch revision so all chains share the same revision
+  uint64 public constant SGHO_REVISION = 2;
```
Kogaroshi: let's set this one to `SGHO_REVISION = 3`, and on Eth mainnet do : 
- upgrade from current Impl to `sGhoInstanceStoragePatch`
- initialize to apply the storage patch
- upgrade from `sGhoInstanceStoragePatch` to this Instance

This way we apply the patch, and finish with a clean implementation, so that the current Eth mainnet deploy looks exactly like other sidechains/L2s in the future. This would be cleaner I think.

Thoughts ? cc @DhairyaSethi

avniculae: for reference, I discussed this with @Kogaroshi and I'm fine to apply the suggested fix. Curious your thoughts @DhairyaSethi

DhairyaSethi: i agree with using 3 here and reinit if using the temp implementation contract in the middle - but tbh im not a fan of using that and am ok with the ugly storage patch logic living in the final canonical implementation contract left after the entire upgrade

avniculae: after disucssions we will use revision 3 such that we do upgrade, fix storage, upgrade and get rid of the extra logic for storage patching

avniculae: we could also simply upgrade to non-patch impl and don't initialize the proxy (if we keep revision 2 then nobody can reinitialise), wdyt? :)

avniculae: in fact I think this is better, as currently the standard init function sets index to 1

tests/fork/sGhoInstanceStoragePatch.fork.t.sol
avniculae: done, the test now fetches supplyCap before the upgrade and checks the migrated value against it

src/contracts/sgho/instances/sGhoInstanceStoragePatch.sol
avniculae: added, `initialize` now reverts with `YieldIndexNotCheckpointed` if `lastUpdate` is not `block.timestamp`, so the migration can't drop accrual from the old rate model. Covered in the fork test.

src/contracts/sgho/instances/sGhoInstance.sol
avniculae: implemented the two-step flow: upgrade to the patch (fix storage) then plain upgrade to sGhoInstance, so no patch logic remains in the final implementation. One note vs the discussion above:
- the patch's `initialize` consumes the shared revision and the final swap passes no calldata. Re-calling `initialize(gho, cap, owner)` on the second upgrade would reset `yieldIndex` to RAY and wipe accrued yield, and leaving the revision unconsumed would let anyone call it and grab `DEFAULT_ADMIN_ROLE`, so the shared revision is what keeps it locked

The fork test mirrors the full flow and checks `initialize` is not callable afterwards.

avniculae: for new instances, initializing at index 1 is not working well given that we need to keep them in sync. will address this with the new feature of delayed rate

Kogaroshi (approved):

src/contracts/sgho/instances/sGhoInstance.sol
avniculae: Went back to revision 2 with a plain final swap, no re-init. Calling initialize on the live vault during the second swap only rewrites values already in storage, and __Pausable_init resets _paused to false, so a vault paused for the upgrade would come out unpaused. New chains still initialize at revision 2 through sGhoInstance (with the checkpoint params from #43), so every chain ends on the same implementation and revision.

PR #43, opened 2026-07-07
feat(sGho): scheduled rate updates & yield index sync for multi-chain deployments

avniculae:
## Summary

Second sGHO feature (on top of #42): make the yield index synchronizable across chains, so sGHO can go multi-chain with the index identical everywhere, at all times.

## Scheduled rate updates

`setTargetRate(uint16 newRate, uint40 effectiveAt)` (`YIELD_MANAGER_ROLE`) is the single rate entry point. The update is stored as pending and applied lazily: views resolve it once due, and the next rate update, deposit/mint or withdrawal/redemption persists it — checkpointing the index **exactly at `effectiveAt`**, never at the chain-specific execution or commit time, so which action (and on which day) happens to persist it on each chain cannot desync the index. Committing on first touch makes the two-segment accrual a one-time cost instead of an overhead on every conversion until the next rate update; plain transfers are exempt, keeping them free of a storage read they otherwise never need.

A multi-chain AIP hardcodes the same `effectiveAt` on every network payload (execution estimate plus a safety delay, e.g. 24h). Payloads execute at different moments per chain, but the rate switches at the same instant everywhere. Note this is deliberately an absolute timestamp rather than a delay stored in the contract: with `now + delay` each chain would flip at its own execution time + delay, so the indexes could never be identical at all times.

Anchoring checkpoints to `effectiveAt` matters beyond the rate switch itself: the accrual division floors, so a checkpoint at any chain-local timestamp would drift the index by a wei. Consequences, documented and covered by tests:
- every scheduled update (even a same-rate one) must be mirrored on all chains;
- passing the current timestamp as `effectiveAt` applies the update immediately, checkpointing at the chain-local execution time — suitable for single-chain deployments only. (There is intentionally no one-arg immediate overload, so this cannot happen by accident; the steward passes `block.timestamp` explicitly.)

The pending schedule packs `effectiveAt` into the free 40 bits of the existing slot (hot path still reads a single slot) and the pending rate into a second slot only touched while a schedule exists. Appending to the ERC-7201 struct leaves the live mainnet layout and the #42 storage patch untouched.

## Cold start & reconciliation

**Cold start via `initialize`**: the initializer now takes the yield index checkpoint (`initialYieldIndex`, `initialLastUpdate`, `initialTargetRate`), so a deployment on a new chain starts in sync directly from a live deployment's checkpoint — the timestamp may be in the past. A genesis deployment passes `(RAY, now, 0)`.

**Reconciliation via `syncYieldIndex(uint120, uint40, uint16)`** (`DEFAULT_ADMIN_ROLE`): overwrites the checkpoint and discards any pending schedule. Since `yieldIndex()`/`lastUpdate()`/`targetRate()` getters resolve a due-but-unpersisted schedule, copying them from an in-sync deployment makes the target chain accrue bit-identically from the checkpoint onward — recovery when the multi-chain AIP fails on one chain (re-schedule any pending update in the same payload).

Both paths share the same validation and `_setCheckpoint` write (`_syncYieldIndex` internal): rate ≤ `MAX_SAFE_RATE`, timestamp not in the future, index ≥ RAY. Sync is admin-gated because it rewrites the index directly (it can lower existing share value, which reconciliation legitimately requires).

## Mainnet upgrade plan (revision 3)

`sGhoInstance.SGHO_REVISION` is now **3** (patch stays at 2). Mainnet flow: checkpoint via the live implementation's legacy `setTargetRate`, upgrade to the storage patch (rev 2, atomic `initialize`), then swap to `sGhoInstance` atomically calling `initialize` with the just-migrated values (a no-op re-affirmation of state). Every chain — mainnet and new deployments — ends locked at revision 3 through the same initializer. The fork test mirrors this exact flow.

## Casting policy

All runtime narrowing casts in `src` and tests go through OZ `SafeCast`. Plain casts remain only where truncation cannot occur or is the point: compile-time-checked literals, and bit-field extraction of packed storage slots (storage patch, test raw-slot decoding).

## Testing

- `TestSGhoTargetRateSchedule`: scheduling, lazy application/persistence, overwrites, accrual around `effectiveAt`, raw-storage laziness checks, and cross-chain determinism tests (incl. fuzz) asserting two deployments with skewed execution times stay bit-identical — this test is what caught the same-rate-reschedule flooring hazard above.
- `TestSGhoSyncYieldIndex`: state/eventing, reverts, pending interaction, share-value effects, cold-start and missed-update reconciliation flows against a live instance, fuzz.
- `TestSGhoInitialization`: cold-start initialize matching a live source chain, initializer validation reverts, revision.
- Fork suite exercises the full three-step mainnet upgrade incl. the final re-initialize.
- Gas snapshots extended; hot-path cost unchanged (deposit +77 gas vs main).
- Full suite: 852 tests pass.


src/contracts/sgho/sGho.sol
```diff
@@ -130,14 +138,39 @@ abstract contract sGho is
...
+      _setCheckpoint(newYieldIndex, block.timestamp.toUint40(), newRate);
+      if (!alreadyCheckpointed) {
+        emit ExchangeRateUpdated(block.timestamp, newYieldIndex);
+      }
+    } else {
+      $.pendingRateEffectiveAt = effectiveAt;
+      $.pendingTargetRate = newRate;
+    }
+    emit TargetRateUpdated(newRate, effectiveAt);
+  }
```
Kogaroshi, line 165: How would that look like if this is called from the Steward on Eth mainnet (`effectiveAt == block.timestamp` then), and needs to be ported to sidechains deployments ? Then we'd have `effectiveAt` in the past ?

avniculae: no, because this would mean between the time this gets updated on mainnet and the time it propagates to sidechains, the sidechains ran on a different rate. They should all be updated with the same future timestamp. The support for past timestamps is not here, but in the sync function, for 2 reasons: new deployments (cold start), and reconciliation in case one sidechain experienced some issues and update with future rate did not go through in a timely manner

src/contracts/sgho/sGho.sol
```diff
@@ -130,14 +138,39 @@ abstract contract sGho is
   }
 
   /// @inheritdoc IsGho
-  function setTargetRate(uint16 newRate) public onlyRole(YIELD_MANAGER_ROLE) {
+  function setTargetRate(uint16 newRate, uint40 effectiveAt) public onlyRole(YIELD_MANAGER_ROLE) {
```
Kogaroshi, line 141: Who is holding that role ? Governance & sGho Steward ?
Because sGho Stewards seems to update to current block timestamp, wouldn't that create some challenges to keep all chains up to date ?

avniculae: only sGhoSteward has the role

wdym by
> Because sGho Stewards seems to update to current block timestamp, wouldn't that create some challenges to keep all chains up to date ?

Kogaroshi: I mean, during this https://github.com/aave-dao/gho-origin/blob/7aa1a44331d32f154f138dcf2cec0577faa01c53/src/contracts/misc/sGhoSteward.sol#L121
I only updates locally, and as the current timestamp. It is not resolved the same across all chains at the same exact timestamp

avniculae: good catch, this is an oversight. The sGhoSteward should have support for future timestamps, fixing

avniculae: fixed in 6b5c6e7: setRateConfig now takes the effectiveAt and passes it through to sGho.setTargetRate, so councils submit the same config with the same effectiveAt on every chain. RateConfigUpdated includes it as well.

src/contracts/sgho/sGho.sol
Kogaroshi: fixed by https://github.com/aave-dao/gho-origin/pull/43#discussion_r3681313666

Kogaroshi (approved):

PR #46, opened 2026-08-03
certora: fix prover timeouts and spec cleanup

avniculae:
Ports Certora/aave-dao-gho-origin#9, which supersedes #6 and #8 — all three can be closed once this lands. Commits and authorship retained; `main` is a direct ancestor, so the diff is only the Certora spec/conf/CI changes.

All 48 prover jobs verify.

On top of the ported work:

- `Rounding.Expand` was modelled as a floor division; the summaries now mirror OZ's `unsignedRoundsUp`
- `mulDiv_equivalence.spec` proves the exact, nondeterministic and closed-form mulDiv encodings denote the same value
- `-copyLoopUnroll 6` was silently dropped on the `optimality.conf R2` line, since a line-level `--prover_args` replaces the conf's array
- `gho-fixedPriceStrategy4626.conf` pointed at a spec that isn't in the tree, and `gho-assetToGhoInvertibility-4626.conf` failed typechecking on a missing `GhoReserve.sol`; both fixed and wired into CI
- `balances-sell-4626 R1/R1a/R2` and `gho-assetToGhoInvertibility-4626` were green only via timeout reruns; both now verify on the committed config, the latter via a z3 seed race (119m timeout → 3m)
- remaining unchecked `twostage` → `twostage-checked`; `optimality4626.spec` moved to the closed-form Ceil encoding to match its non-4626 twin

One pre-existing item left open: `getAmount-properties-4626.conf --rule getAssetAmountForSellAsset_optimality getAssetAmountForBuyAsset_funcProperty` has gone green via rerun on two of five runs (#issuecomment-5177477733).


.github/workflows/certora-gsm-4626.yml
```diff
@@ -58,7 +58,9 @@ jobs:
...
             certora/gsm/conf/gsm4626/finishedRules-4626.conf --rule giftingUnderlyingDoesntCreateExcessOrDearth
-            certora/gsm/conf/gsm4626/gho-assetToGhoInvertibility-4626.conf
+            # One rule per line: both are heavy enough that running them together times out.
+            certora/gsm/conf/gsm4626/gho-assetToGhoInvertibility-4626.conf --rule buyAssetInverse_all
+            certora/gsm/conf/gsm4626/gho-assetToGhoInvertibility-4626.conf --rule sellAssetInverse_all
```
shellygr: are you sure we cover everything in the spec file for this conf? that's why we sometimes do `--rule R` and then `--exclude_rule R`

avniculae: Good catch — you were right that enumerating rules with `--rule` is fragile: add a third rule to that spec and it would silently never run with CI still green.

In the end the split went away entirely rather than being converted to `--exclude_rule`. The timeout that made me split it was fixable with the z3 seed race (see follow-up below), so the workflow is back to a single bare `gho-assetToGhoInvertibility-4626.conf` line with no rule enumeration at all.

Worth knowing this conf had never run in any workflow: its `files` list was missing `GhoReserve.sol` that `methods4626_base.spec` needs, so it failed typechecking.

*(Edited: originally said I had excluded `sellAssetInverse_all` and asked for a ticket number to reference it. Both no longer apply — nothing is excluded and no ticket is needed.)*

avniculae: Update — I was wrong to exclude `sellAssetInverse_all`, and the fix was your own trick sitting in four sibling confs.

The assert is provable; it was just landing near the timeout. Adding the z3 random-seed race that `balances-buy`, `fees-buy`, `fees-buy-4626` and `optimality4626` already use:

| | before | after |
|---|---|---|
| `sellAssetInverse_all` alone | TIMEOUT at 119m (5 of 6 asserts) | verified, 3m, 6 asserts |
| whole conf, both rules | 67m | verified, 5m, 11 asserts |

So no rule is dropped, no ticket needed, and I've put the flag in the conf's `prover_args` rather than on the workflow line — both because that's where the other four keep it, and because a `--prover_args` on the line would replace the conf's array rather than add to it.

That also removes the thing you flagged: with the conf fast again there's no reason to split at all, so the workflow is back to a single bare `gho-assetToGhoInvertibility-4626.conf` line with no rule enumeration to go stale.

What led me astray was treating one unlucky 119m run as evidence the rule couldn't converge. Ignore the earlier ticket-number request.

Kogaroshi (approved):

.github/workflows/certora-gsm-4626.yml
```diff
@@ -27,19 +27,26 @@ jobs:
...
 
-      - uses: Certora/certora-run-action@v2
+      - uses: Certora/certora-run-action@v2.10.0
         with:
-          cli-version: 7.31.2
+          cli-release: stable
+          cli-version: "8.17.1"
+          auto-rerun-timeouts: "true"
+          # NB: a `--prover_args` here REPLACES the conf's `prover_args` array, it does not
+          # append to it. Restate every flag the run needs, not just the one being added.
```
DhairyaSethi: do we need this comment here?

.github/workflows/certora-gsm-4626.yml
```diff
@@ -27,19 +27,26 @@ jobs:
         with:
           submodules: recursive
 
-      - uses: Certora/certora-run-action@v2
+      - uses: Certora/certora-run-action@v2.10.0
```
DhairyaSethi: we've been pinning by hash usually bc this can be updated as well but nbd this is fine also

certora/gho/specs/ghoToken.spec
```diff
@@ -140,7 +131,6 @@ invariant address_in_set_values_iff_in_set_indexes(address facilitator)
 	is_in_facilitator_set_array(facilitator) <=> is_in_facilitator_set_map(facilitator)
 	{preserved{
 		requireInvariant addressSetInvariant_2();
-		requireInvariant length_leq_max_uint160();
```
DhairyaSethi, line 143: why do we not need this anymore?

certora/gsm/specs/shared/mulDiv_summaries.spec
```diff
@@ -0,0 +1,52 @@
...
+        return mulDivNondet(x, y, denominator);
+    }
+    require denominator > 0;
+    uint256 q;
+    require x * y <= q * denominator;
+    require q * denominator < x * y + denominator;
+    return q;
+}
+// Same value as mulDivRounding, written as the single closed form (x*y+d-1)/d for Ceil.
+// The optimality proofs reason about this closed form far better than the reuse-floor form.
```
DhairyaSethi, line 45: very interesting, curious if it's bc of the consecutive random number gen (seed) 

need to look into how that z3 rand works bc if it's deterministic then symbolically we're not checking all cases and instead adding a restriction correlation

DhairyaSethi (approved):

certora/gsm/specs/shared/mulDiv_summaries.spec
avniculae: what do you mean here? what cases are we leaving out here?

certora/gho/specs/ghoToken.spec
avniculae: `set.spec:109` has `axiom to_mathint(mirrorArrayLen) < TWO_TO_160() - 1` unconditionally, and the array-length Sload hook forces `mirrorArrayLen == len`, so the bound already holds without the `requireInvariant`.

The invariant itself was vacuous — its own docstring said so, "the proof of the assumption is vacuous because length > loop_iter" — so it went along with its five uses. `verifyGhoToken` went 29 -> 28 rules, just that one.

.github/workflows/certora-gsm-4626.yml
avniculae: Pinned by SHA across all four workflows in d63123f.

.github/workflows/certora-gsm-4626.yml
avniculae: Dropped in d63123f, from both workflows that had it.

